PR #11, opened 2022-08-17
Updated dockerfile & added release workflow

### LINKED ISSUE #6: Provide Dockerfile as basis for testing in this repository
Related to #3. Currently we are using a specific development version of FEniCS-X for our development. We should add a Dockerfile to this repository which is used as a basis for testing. A good template is provided here: https://github.com/nemocrys/dolfinx-openfoam/blob/main/docker/Dockerfile.

We need the follwoing:

* FEniCS-X (we now have to build manually, ~~but should be replaced with `dolfinx/dolfinx:latest` in the future.~~ but should be replaced with installation of a binary in the future; see [the dockerfile for the fenics-adapter](https://github.com/precice/fenics-adapter/blob/develop/tools/releasing/packaging/docker/Dockerfile))
* preCICE (use preCICE 2.3.0)
* python-bindings (use 2.3.0.1)

We don't need/want:

* OpenFOAM and other dependencies.

boris-martin:
Changed the dockerfile to install FEniCSx from Ubuntu packages, then install the adapter. I also added a Github Action to push an image on Docker Hub. It would work once 1) A token is set on the repository 2) The PR #1 is merged. Before pushing, I added a rudimentary test: push if and only if `python3 -c "import fenicsxprecice"` runs without error.

Currently, the adapter is installed in regular mode. It might be interesting to add the `-e` flag so that this image can be used efficiently for debugging. Opinions welcome!


IshaanDesai (approved):

arvedes (approved):

boris-martin:
So when you start the container you remove the pre-built with `pip uninstall` then install the mounted one? Or is it enough to install the new one because pip will override the pre-installed version?
I like your workflow (I lost some time setting up remote editing through Docker, adding Git credential into new containers etc) so it makes sense, but what's the purpose of pre-installing anything then? Wouldn't you want just a container with FEniCSx and no adapter at this point? Shall we just keep a pre-installed version only for automated testing ?

arvedes:
I used a container without adapter: https://github.com/nemocrys/dolfinx-openfoam/blob/main/docker/Dockerfile . I'm not sure about the behavior with pre-installed adapter, I'd just try it out. 

The version with the pre-installed adapter is definitely useful for testing. Maybe, some people will use it as a reference for installation (at least I do that sometimes ;) ) or directly work with the container in their own projects.

boris-martin:
I just tested, and installing local version automatically triggers a (clean) uninstall of the pre-installed version. 
So let's keep the non-editable install, then the user will be free to override it easily.

Issue #12, opened 2022-08-18
Linter complains about dead links due to non-existent releases


boris-martin:
The Markdown Linter flow fails because of links to non-existent things such as https://pypi.org/project/fenicsxprecice/#history
Should we release at some point to make it existent, update the doc to remove dead links, or simply ignore the issue until the adapter is ready ?

IshaanDesai:
We do want to release in the near future, but till then we can remove or temporarily comment out such links so that the markdown linter workflow passes. I understand that is annoying to have a failing workflow when one is trying to merge PRs.

Issue #9, opened 2022-04-29
Enable Vector valued coupling


arvedes:
Currently, the adapter is only working with scalar-valued data. For vector valued data there is currently a "not tested" error.

BenjaminRodenberg:
In #15 we removed support for vector valued data. We are now raising an exception. Adding support for vector valued data is still on the agenda but not necessary for our first, simple test (partitioned heat equation).

Issue #2, opened 2022-02-24
Develop existing FEniCS tutorials here


BenjaminRodenberg:
To make development easier we develop the FEniCS-X tutorials here. As soon as this adapter is stable, we will migrate the tutorials to https://github.com/precice/tutorials. Our goal is to port all FEniCS tutorials to FEniCS-X. This is the complete list:

* [`partitioned-heat-conduction`](https://github.com/precice/tutorials/tree/master/partitioned-heat-conduction): As a test for simple geometries and scalar valued coupling data
* [`partitioned-heat-conduction-complex`](https://github.com/precice/tutorials/tree/master/partitioned-heat-conduction-complex): As a test for complex geometries and vector valued coupling data.
* [`volume-coupled-diffusion`](https://github.com/precice/tutorials/tree/master/volume-coupled-diffusion): As a test case for volume coupling
* [`perpendicular-flap`](https://github.com/precice/tutorials/tree/master/perpendicular-flap): For FSI
* [`flow-over-heated`](https://github.com/precice/tutorials/tree/master/flow-over-heated-plate): For CHT

We should follow the folder structure and naming of the `tutorials` repository and put everything into the folder `fenicsx-adapter/tutorials`. The second participant (openfoam, su2...) and the preCICE configuration file should not be pushed into this repository, but just manually copied from the tutorials.

BenjaminRodenberg:
#15 and https://github.com/precice/tutorials/pull/317 represent a stable case for a limited set of features. We therefore decided to properly integrate the tutorial in the tutorials repository.

IshaanDesai:
Now that we have ported the `partitioned-heat-conduction` tutorial from this repo into the tutorials, do we still want to pursue the strategy of developing tutorials here and porting them later?

BenjaminRodenberg:
From my current perspective: No. I think the API is sufficiently stable that we should be able to develop the remaining tutorials directly in the tutorials repository. If we find out that we run into problems, we can just reopen this issue. We can still create an issue in the tutorials repo to keep track of the tutorials that still need to be ported. This is also a nice overview, if somebody wants to do a thesis or HiWi project on porting one of the tutorials.

IshaanDesai:
Agreeing with @BenjaminRodenberg and hence closing this issue.

PR #15, opened 2022-12-14
Develop minimal working version without support for parallel runs and vector-valued functions

PhilipHildebrand:
This PR is for merging the fixed error computation of the partitioned-heat-equation tutorial into develop. Furthermore, any tests of the adapter regarding vector functions have been removed, because vector functions are currently not working. 
Important to note is that executing the tests via `python3 setup.py test` fails for some reason; executing the tests/ test suites individually does work however.


fenicsxprecice/expression_core.py
```diff
@@ -200,49 +166,28 @@ def create_interpolant(self):
...
+                # TODO check if self._vals[:, d] is required here, above it had to be removed
+                raise Exception("Not tested")
                 interpolant.append(self.segregated_interpolant_2d(self._coords_x, self._coords_y, self._vals[:, d]))
         else:
             raise Exception("Problem dimension and data dimension not matching.")
 
         return interpolant
 
-    def interpolate(self, x):
+    def interpolate_precice(self, x):
```
BenjaminRodenberg, line 181: See above.

fenicsxprecice/expression_core.py
```diff
@@ -55,11 +56,11 @@ def update_boundary_data(self, vals, coords_x, coords_y=None, coords_z=None):
         self._f = self.create_interpolant()
 
         if self.is_scalar_valued():
-            assert (self._vals.shape == self._coords_x.shape)
+            assert (self._vals.shape[0] == self._coords_x.shape[0])
         elif self.is_vector_valued():
             assert (self._vals.shape[0] == self._coords_x.shape[0])
 
-    def interpolate(self, x):
+    def interpolate_precice(self, x):
```
BenjaminRodenberg, line 63: Why was this renamed? If you can use the old name, I would suggest to keep it to reduce the diff.

fenicsxprecice/adapter_core.py
```diff
@@ -29,30 +27,23 @@ class VertexType(Enum):
 
 class Vertices:
     """
-    Vertices class provides a generic skeleton for vertices. A set of vertices has a set of global IDs, local IDs and
+    Vertices class provides a generic skeleton for vertices. A set of vertices has a set of IDs and
     coordinates as defined in FEniCSx.
     """
 
     def __init__(self, vertex_type):
         self._vertex_type = vertex_type
```
BenjaminRodenberg, line 35: Is the vertex type still needed? I only found vertices of type `FENICSX` (did not look extremely carefully though). If all vertices are of this type I don't see a clear reason for specifying the type and we should remove it.

fenicsxprecice/adapter_core.py
```diff
@@ -160,17 +120,24 @@ def convert_fenicsx_to_precice(fenicsx_function, local_ids):
...
+    x_mesh = fenicsx_function.function_space.mesh.geometry.x
+    x_dofs = fenicsx_function.function_space.tabulate_dof_coordinates()
+    mask = []  # where dof coordinate == mesh coordinate
+    for i in range(x_dofs.shape[0]):
+        for j in range(x_mesh.shape[0]):
+            if np.allclose(x_dofs[i, :], x_mesh[j, :], 1e-15):
+                mask.append(i)
+                break
+    sampled_data = fenicsx_function.x.array[mask]
+    # end dirty fix
```
BenjaminRodenberg, line 134: I know we already discussed this. Why is this dirty fix needed? I understand if for the current situation it is good enough, but I think it would be good to open an issue that explains what's happening here, why a dirty fix is necessary and how a clean alternative could look like. @PhilipHildebrand could you take care of opening the issue and put a link to the issue here into the comments? I think @arvedes and I can try to add some details in the issue then. (For the sake of this PR the dirty fix is sufficient)

fenicsxprecice/adapter_core.py
```diff
@@ -160,17 +120,24 @@ def convert_fenicsx_to_precice(fenicsx_function, local_ids):
...
 
     if len(local_ids):
-        if fenicsx_function.function_space().num_sub_spaces() > 0:  # function space is VectorFunctionSpace
-            for lid in local_ids:
-                precice_data.append(sampled_data[:, lid])
+        if fenicsx_function.function_space.num_sub_spaces > 0:  # function space is VectorFunctionSpace
+            raise Exception("Functions from VectorFunctionSpaces are currently not supported.")
+            #for lid in local_ids:  
+               #precice_data.append(sampled_data[lid, :])  # this line raises an IndexError
```
BenjaminRodenberg, line 140: Please delete the commented out lines of code. Please also link to https://github.com/precice/fenicsx-adapter/issues/9 in the exception.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -62,52 +62,28 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
         self._dofmap = None  # initialized later using function space provided by user
 
         # coupling mesh related quantities
-        self._owned_vertices = Vertices(VertexType.OWNED)
-        self._unowned_vertices = Vertices(VertexType.UNOWNED)
         self._fenicsx_vertices = Vertices(VertexType.FENICSX)
```
BenjaminRodenberg, line 65: See comment above: I think we don't need any type for the vertices, if there are only vertices of `VertexType.FENICSX`.

README.md
```diff
@@ -20,10 +20,11 @@ Notes:
 
 * This adapter is a fork from the original [FEniCS-preCICE adapter](https://github.com/precice/fenics-adapter). Based on [v1.2.0](https://github.com/precice/fenics-adapter/releases/tag/v1.2.0).
 * This adapter is currently under development and experimental.
+* Target version: dolfinx v0.4.1
```
BenjaminRodenberg, line 23: Just to make sure: Is this the version you are using, as well?

BenjaminRodenberg (changes requested):
Thanks for opening this PR! This is a big step into the direction of getting the adapter working again for a minimal test case. I added some comments below. I think we should try to make this really a minimal working example. The less code we have in the end, the better (as long as it is working with our test case, of course).

Additionally: Can you please [open a PR in the tutorials repository](https://github.com/precice/tutorials/compare) and move everything from `fenicsx-adapter/tutorials/partitioned-heat-condiction` to the correct location in the tutorials repo? Our aim should be a proper intergation of the example into the tutorials repository. Related to https://github.com/precice/fenicsx-adapter/issues/2, feel free to put a brief comment into that issue.

Additionally, some minor documentation todos:

* Please open an issue that we do not forget to fix running the tests via `python3 setup.py test`. Btw: I just discovered that the tests actually work correctly with our CI pipeline (https://github.com/precice/fenicsx-adapter/pull/15/checks). That's good news. Maybe this gives you a hint for debugging the way you are running the tests locally. If I find some time I can also try to run the tests on my system to see what's happening.
* Please change the title of this PR. Not "Merge ..." (clear that you are merging, because it is a PR), but what you did here. "Develop minimal working version without support for parallel runs and vector-valued functions."
* Use a linter to autoformat the code. Currently the corresponding test is failing (https://github.com/precice/fenicsx-adapter/actions/runs/3696371324/jobs/6259976119). You find the right commands for the linter in the corresponding test pipeline: https://github.com/precice/fenicsx-adapter/blob/develop/.github/workflows/check-pep8.yml

fenicsxprecice/adapter_core.py
PhilipHildebrand: You're right, the vertex type can be removed.

fenicsxprecice/adapter_core.py
PhilipHildebrand: Sure, here is the link: [https://github.com/precice/fenicsx-adapter/issues/17](url)

README.md
arvedes: Is it possible to go for v0.5.2? I don't think there have been (m)any breaking changes, shouldn't be a big deal to update.

BenjaminRodenberg:
Some of the problems with the linter are originating from dead links. This is related to https://github.com/precice/fenicsx-adapter/issues/12 and we should fix this directly on `develop`. I will quickly create a corresponding PR.

BenjaminRodenberg:
> Some of the problems with the linter are originating from dead links. This is related to #12 and we should fix this directly on `develop`. I will quickly create a corresponding PR.

I took care about this in #18 and synchronized this PR to develop. The links should be working now.

fenicsxprecice/adapter_core.py
BenjaminRodenberg:
```suggestion
    # TODO begin dirty fix. See https://github.com/precice/fenicsx-adapter/issues/17 for details.
    x_mesh = fenicsx_function.function_space.mesh.geometry.x
    x_dofs = fenicsx_function.function_space.tabulate_dof_coordinates()
    mask = []  # where dof coordinate == mesh coordinate
    for i in range(x_dofs.shape[0]):
        for j in range(x_mesh.shape[0]):
            if np.allclose(x_dofs[i, :], x_mesh[j, :], 1e-15):
                mask.append(i)
                break
    sampled_data = fenicsx_function.x.array[mask]
    # end dirty fix
```

README.md
BenjaminRodenberg: On my local system I just installed `v0.5.2` via `apt`. I will see, if this causes any problems and fix them, if necessary.

BenjaminRodenberg: Everything works nicely with `v0.5.2` (ran tests + partitioned heat equation tutorial). I'm updating the target version now via https://github.com/precice/fenicsx-adapter/pull/20.

fenicsxprecice/expression_core.py
BenjaminRodenberg: I think I know what's the problem now: `dolfinx.fem.Function` already has a function `dolfinx.fem.Function.interpolate`. This function is probably needed by the class. Therefore, we are not allowed to override it. I'm still not a big fan of the name `interpolate_precice`. Maybe we can get rid of the function `interpolate_precice` entirely and put everything into `update_boundary_data`. I will try to get this working.

BenjaminRodenberg: I managed to simplify the implementation in c8e4b75a9ae71104ffef81e662e458149cb7a0dd. It was quite convoluted before and now it is much nicer (because of the way how FEniCS-X allows us to create an interpolant). @PhilipHildebrand: Please take a look at c8e4b75a9ae71104ffef81e662e458149cb7a0dd. We can also discuss the changes together.

fenicsxprecice/expression_core.py
BenjaminRodenberg: See [c8e4b75](https://github.com/precice/fenicsx-adapter/pull/15/commits/c8e4b75a9ae71104ffef81e662e458149cb7a0dd)

fenicsxprecice/adapter_core.py
BenjaminRodenberg: I took care of this in [9e6510a](https://github.com/precice/fenicsx-adapter/pull/15/commits/9e6510aad549a59cf3e57d5d94173e0bbd647638).

BenjaminRodenberg (changes requested):
I think we are close to done here. @PhilipHildebrand can you take care of the unneeded vertex types? I will then do a final review and merge.

fenicsxprecice/adapter_core.py
BenjaminRodenberg: This was fixed in [0d2f727](https://github.com/precice/fenicsx-adapter/pull/15/commits/0d2f7278b3b6880163c602e31f1c06ce21edb196).

fenicsxprecice/fenicsxprecice.py
BenjaminRodenberg: Also [0d2f727](https://github.com/precice/fenicsx-adapter/pull/15/commits/0d2f7278b3b6880163c602e31f1c06ce21edb196).

fenicsxprecice/adapter_core.py
```diff
@@ -196,445 +147,34 @@ def get_fenicsx_vertices(function_space, coupling_subdomain, dims):
...
-    for recv_gid in recv_pts.keys():
-        recv_lid = int(np.where(fenicsx_gids == recv_gid)[0][0])
-        coupling_data[tuple(fenicsx_coords[recv_lid])] = recv_reqs[counter].wait()
-        counter += 1
-
-    return coupling_data
+            coords = mesh.geometry.x[ids][:, :2]
+        else:
+            coords = np.array([])
+    except Exception as e:  # fall back to old method  # TODO is that to general? Better use, e.g., IndexError here?
```
BenjaminRodenberg:
```suggestion
    except Exception as e:  # fall back to old method  # TODO is that too general? Better use, e.g., IndexError here?
```

BenjaminRodenberg (approved):
Looks good. There are issues for the open points, namely parallelization #22, minor updates in the CI #21, removing the dirty fix #17 and allow vector functions #9.

PR #20, opened 2022-12-16
Move to target version v0.5.2.

BenjaminRodenberg:
Move to target version v0.5.2 for mock unit tests.


BenjaminRodenberg:
Tests are failing due to compatibility problems that will be solved via #15.

Issue #17, opened 2022-12-15
Cleaning the dirty fix in `adapter_core.py`


PhilipHildebrand:
Right now, there is a dirty fix implemented for collecting the sampled data. Is there a particular reason for this?

*edit by @BenjaminRodenberg*

https://github.com/precice/fenicsx-adapter/blob/26678c169022bfc1776f1fa2d8bad40e4fbf825b/fenicsxprecice/adapter_core.py#L124-L134

arvedes:
I implemented it when updating to v0.4, because the property `fenicsx_function.x.array` we used before was removed. See also:
https://github.com/precice/fenicsx-adapter/pull/1#issuecomment-1163222926
https://github.com/precice/fenicsx-adapter/pull/1/commits/5af2065c8b8e4c53edcb877dee17e80724ab19a4

PR #25, opened 2023-01-23
Moved mask computation into adapter initialization

PhilipHildebrand:
This PR is to move the mask computation into the initialization of the adapter. This also resolves the [issue](https://github.com/precice/fenicsx-adapter/issues/17) about the dirty fix.

### DIFF
--- a/fenicsxprecice/adapter_core.py
+++ b/fenicsxprecice/adapter_core.py
@@ -85,7 +85,7 @@ def determine_function_type(input_obj):
         raise Exception("Error determining type of given dolfin FunctionSpace")
 
 
-def convert_fenicsx_to_precice(fenicsx_function, local_ids):
+def convert_fenicsx_to_precice(fenicsx_function, mask, local_ids):
     """
     Converts data of type dolfin.Function into Numpy array for all x and y coordinates on the boundary.
 
@@ -93,6 +93,8 @@ def convert_fenicsx_to_precice(fenicsx_function, local_ids):
     ----------
     fenicsx_function : FEniCSx function
         A FEniCSx function referring to a physical variable in the problem.
+    mask : numpy array
+        Array of Function degrees of freedom on which the data gets sampled
     local_ids: numpy array
         Array of local indices of vertices on the coupling interface and owned by this rank.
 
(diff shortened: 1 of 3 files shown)


PhilipHildebrand:
I've added a figure that shows the time improvement for the partitioned heat equation tutorial (for ny = 9, 18, 36, 72 and 144 respectively). For ny = 72 and ny = 144, the version with the dirty fix in it still wasn't finished after several hours , so I've omitted those cases.
[timeEval-1.pdf](https://github.com/precice/fenicsx-adapter/files/10539120/timeEval-1.pdf)

IshaanDesai:
You can add the figure by directly dropping it in the comment while you are writing it. That would make the figure visible inside the comment. I am surprised to know that for `ny = 72` or `144`, we already have saturation. Around a hundred nodes along the coupling interface is not a lot. Can you investigate on which part of the adapter is taking long? Is it the mask computation again?

IshaanDesai:
Also, please make sure that all tests pass. Currently one tutorial seems to fail and there is also an error from the formatter.

PhilipHildebrand:
> You can add the figure by directly dropping it in the comment while you are writing it. That would make the figure visible inside the comment. I am surprised to know that for `ny = 72` or `144`, we already have saturation. Around a hundred nodes along the coupling interface is not a lot. Can you investigate on which part of the adapter is taking long? Is it the mask computation again?

Actually, the adapter is slower all around, not just the mask computation. For reference, for `ny = 72`,  roughly 51% of the runtime are spent during the mask computation, for `ny = 144`it is roughly 63%.

IshaanDesai:
> Actually, the adapter is slower all around, not just the mask computation. For reference, for `ny = 72`, roughly 51% of the runtime are spent during the mask computation, for `ny = 144`it is roughly 63%.

How much slower is the adapter as compared to the FEniCS adapter? This would be a good thing to find out. Also, is FEniCSx slower than FEniCS or is our adapter slower?

PhilipHildebrand:
> > Actually, the adapter is slower all around, not just the mask computation. For reference, for `ny = 72`, roughly 51% of the runtime are spent during the mask computation, for `ny = 144`it is roughly 63%.
> 
> How much slower is the adapter as compared to the FEniCS adapter? This would be a good thing to find out. Also, is FEniCSx slower than FEniCS or is our adapter slower?

I tested the FEniCS tutorial using the [preCICE VM](https://precice.org/installation-vm.html) and it is performs way better for high `ny`, at `ny = 144` it doesn't exceed 12 s. However, I don't know whether it's because of our adapter or because FEniCSx is slower than FEniCS in general.
Fenics


FenicsVsFenicsx

IshaanDesai:
> However, I don't know whether it's because of our adapter or because FEniCSx is slower than FEniCS in general.

It should be straightforward to find this out. Using functionality in Python to measure time, you can get how much time is being spent in adapter functions and also how much time the solver is taking. It would be a good insight to find out if the solvers are the issue or is it something on our side.

NiklasVin:
This PR can be closed as the changes regarding the dirty fix are outdated since PR https://github.com/precice/fenicsx-adapter/pull/28 as it was removed.

PR #26, opened 2024-07-16
Adapter precice 3.1 with 3D support and "point load"

efirvida:
Sorry If I dont follow the source style here, my point was to quick develop a solution when it works, we can do it well. This also can keep the code more clean, to discuss over it. Also I have to say that I'm learing preCICE and FenicsX at the same time, so maybe the solution here will be simple and I can't notice it.

I post a first version of the adapter on the [preCICE Disource](https://precice.discourse.group/t/fsi-with-fenicsx-and-precice-3/1966) because at that point I have where I call a scaling problem in the flap tutorial, having similar results but with 1 to100 diference in the result.  

Then I create a simple static beam problem comparing the results of CaluliX and FenicsX and here I notice the problem, I was integrating the load over the surface, instead of on a DoF as Calculix does, which was confirmed by [BenjaminRodenberg](https://precice.discourse.group/u/BenjaminRodenberg) in the discourse. This is the example that I put  on the test folder. So I update my the flap tutorial based on the lessons learned here.

This also guided by [this other thread on FeniscX forum](https://fenicsproject.discourse.group/t/improve-mapping-data-on-a-function-by-coordinates/14911/9), about mapping the loads over DoFs. But I still having problems validting the flap tutorial. Considering that the FenicsX dos'nt have a PointSource as the legacy version, I use another approach of modifiying directly the vector value as:

```
LOAD_POINT = (4.9490529e-02, -1.4846031e-01, 1.5000000e02)
LOAD = np.array([10.0, 0, 0])

def point(x):
    return (
        np.isclose(x[0], LOAD_POINT[0])
        & np.isclose(x[1], LOAD_POINT[1])
        & np.isclose(x[2], LOAD_POINT[2])
    )

# https://bleyerj.github.io/comet-fenicsx/tours/dynamics/elastodynamics_newmark/elastodynamics_newmark.html
point_dof = fem.locate_dofs_geometrical(V, point)[0]
point_dofs = np.arange(point_dof * dim, (point_dof + 1) * dim)

b_func = fem.Function(V)
b_func.x.array[:] = 0
b = b_func.vector
with b.localForm() as b_local:
    b_local.set(0)

for i, j in enumerate(point_dofs.tolist()):
    b[j] = LOAD[i]
b.assemble()

```
Which works for a single point, but I'm missing somthing here, and I dont know if is on the preCICE side or in the implementation of the flap tutorial. Rigth now I can confirm that the adapter can handle 2D and 3D problems, but without good results. 

1. https://fenicsproject.discourse.group/t/improve-mapping-data-on-a-function-by-coordinates/14911/9
2. https://precice.discourse.group/t/fsi-with-fenicsx-and-precice-3/1966


efirvida:
I finally made it work, at least the perpendicular flap tutorial, To do it I added a `LinearDiscreteProblem` that puts the load directly on the DoFs and also maintains the code clean. 

flap-x-displacement-tip


I added two more examples the turek-heron, and a turbine-tower, right now I'm running them to see what happens. I'm having problems by running the turbine-tower in parallel from the OpenFOAM side, but I don't know how to diagnose what happen

IshaanDesai (commented):
Thank you for your contribution and apologies for the delay in reviewing. Even if the total number of new files is large, as a first step I looked at the adapter files, so the folder `fenicsxprecice`. I see that you have broadly stuck to the existing adapter class structure, by having a `Config` class, an `Adapter` class, and a class to manage the state for checkpointing. To me it seems like this could have potentially been accommodated in the existing adapter structure. What was the reason to completely discard the existing adapter and write a new one?

Contents of the file `fenicsxprecice/solvers.py` is related to the numerical solvers. Such code typically goes in the application files, as it is not directly related to coupling.

Apart from the adapter files, there are a lot of files related to different simulation setups. Ideally such simulation setups are added to the [precice/tutorials](https://github.com/precice/tutorials) repository, rather than the adapter repository. Once we start to converge on the adapter contribution, we can move the case files to the tutorials repository.

efirvida:
@IshaanDesai thanks for your quick review. Yes, I use the original adapter structure, because first I tried to use the original on my FSI by doing the small modifications. As I'm new on preCICE and DOLFINx, I began to modify the code in place to do some reverse engineering, until I reached good results, and I removed a lot of code that I didn't understand to keep my version as clean as possible with the minimum needs to perform an FSI, as you can see the API on this version is a lot simpler than the original. 

I know that I put a lot of tutorials that do not belong here, as de solver.py, but I think it is easier to have all things together in one place to test/review before moving/merge and put all in the right place. But OK I did some cleaning 

The main part of the Adapter are the methods:

```
interpolation_points_in_vector_space
interface_dof
```
the first return the DOFs and coordinates of the interface. and then is easy to use the `interface_dof` to read and write data from the evaluated Function. I tried to add another tutorial on heat transfer but I've not time to do it to see if this approach also works for this kind of solution. Also, the Turek-FSI isn't working well I think because this is a nonlinear problem and I use a linear solver.

IshaanDesai:
> Yes, I use the original adapter structure, because first I tried to use the original on my FSI by doing the small modifications. As I'm new on preCICE and DOLFINx, I began to modify the code in place to do some reverse engineering, until I reached good results, and I removed a lot of code that I didn't understand to keep my version as clean as possible with the minimum needs to perform an FSI, as you can see the API on this version is a lot simpler than the original.

I understand how your reached the state you have, and that it serves the purpose of doing FSI simulations. But the original structure is intended to be for generic use and was also designed rigorously for legacy FEniCS. I am tending towards keeping the old structure and accommodating your improvements in it. Did you encounter any fundamental design problems while trying to incorporate your implementation in the current adapter? If yes, we can improve the design.

> I know that I put a lot of tutorials that do not belong here, as de solver.py, but I think it is easier to have all things together in one place to test/review before moving/merge and put all in the right place. But OK I did some cleaning

This is perfectly fine for now, but just something to do at the end, just before merging.

> I tried to add another tutorial on heat transfer but I've not time to do it to see if this approach also works for this kind of solution.

This would be very interesting to know. If this can be done, we could potentially just move to the new design.

> Also, the Turek-FSI isn't working well I think because this is a nonlinear problem and I use a linear solver.

This is a known problem.

I would suggest the following next steps:

1. I can help you to try to incorporate your implementation into the current adapter design. This way we can keep the design but you can also do FSI simulations.
2. Move your simulation setups to the tutorials repository one by one.
3. If possible, add some tests for the new functionality.

What do you think? I will, of course, assist to a large degree here.

efirvida:
I can try to adapt my implementation into the current adapter, but for this, I would need your help, I don't remember the problems in the main implementation.

I think that the main cut-off is on the `initialize` method where I believe the `write_object` is not needed here, or at least I didn't understand its function. Maybe due to the simple implementation of the read/write data methods, that I put the data directly on the Function.

And the same for the `create/update_coupling_expression` that I think we don't need here, at least for FSI.

Right now this version has a problem working in parallel, and I'm working on it right now, the I could try to start mergin with the current implementation. Notice that the current implementation also is for preCICE < 3, and this implementation not, so here some method changes its names and parameters.

IshaanDesai:
> I can try to adapt my implementation into the current adapter, but for this, I would need your help, I don't remember the problems in the main implementation.

I can and will definitely help here.

I see that you extract the data from the function directly at the time of reading and writing. I will check how this would work with the original implementation.

Handling FEniCS Expressions was part of the original design because many times Expressions were used on the user-side. We can rethink this now.

> Right now this version has a problem working in parallel, and I'm working on it right now, the I could try to start mergin with the current implementation.

I saw the Discourse post about the issue with parallel runs. Let me try to answer there, so as to not clutter this pull request too much. Thanks for agreeing to merging with the current implementation :smile: I know it feels like taking a step backwards, but in the long run we would benefit from the original design decisions.

> Notice that the current implementation also is for preCICE < 3, and this implementation not, so here some method changes its names and parameters.

Thanks for pointing this out :+1:

efirvida:
I accidentally overwrote my last commit, but I believe it's not a significant issue. I modified the adapter using the current class reference to simplify the merging process.

In the current implementation, the read and write methods are much simpler because preCICE v3 does not differentiate between vectors and scalars, using the same function for both. I also removed some checks that I didn't fully understand, and I'm not certain if they were necessary.

```
        assert (self._coupling_type is CouplingMode.UNI_DIRECTIONAL_READ_COUPLING or
                CouplingMode.BI_DIRECTIONAL_COUPLING) 
```

As I mention in my last post, I'm porting the adapter to preCICEv3 so some methods change, now we have:

- `requires_initial_data` instead of `is_action_required`
- `requires_reading_checkpoint` instead of `action_read_iteration_checkpoint`
- `requires_writing_checkpoint` instead of `action_write_iteration_checkpoint`

I also added some new properties:

- `dt`: This property captures the participant's time step (dt), so it's no longer necessary to specify it separately in our problem setup.
- `interface_dof`: This is the most important property, enabling us to read and write interface data. You can think of it as analogous to the PointSource in the legacy FEniCS version.
- `interface_coordinates`: This property is mainly for debugging purposes. It allows us to visualize the interface interpolation points by saving them as a CSV file, which can then be plotted.

So the API change a little bit due to this new properties, and the change of the preCICE version.

Also in the `initialize` method I left the `write_object` param just to maintain compatibility, but my current implementation don't need it. And also I remove all the check on this methods because I didn't understand, and I'm not sure if they were necessary.

FinallyI include the flap tutorial again with this modification, jus to show how we can use it.

efirvida:
Also I manage to add a working heat transfer tutorial, in [this commit](https://github.com/efirvida/fenicsx-adapter/commit/d6c5320dddece98460a1b24a7670523c6b884f05#diff-e1b5c6d7ff07eb6e9b3518ee669e964174430d22c918fd2cf564378219943a64) of my main branch (which is not aligned right now with this last version). The tutorial  runs until the end and change data between participants, but as I'm not an expert in heat transfer I'm sure that there are things wrong there on the FEniCSx side implementation, but it's a starting point

 LICENSE
```diff
@@ -162,4 +162,4 @@ General Public License ever published by the Free Software Foundation.
 whether future versions of the GNU Lesser General Public License shall
 apply, that proxy's public statement of acceptance of any version is
 permanent authorization for you to choose that version for the
-Library.
+Library.
```
IshaanDesai: Lets not change the license file. Can you please keep the original file?

.gitattributes
```diff
@@ -1 +1 @@
-fenicsxprecice/_version.py export-subst
```
IshaanDesai: Same as the license file.

.github/workflows/build-and-test.yml
IshaanDesai, line 1: We should keep the build and test CI, hence we need this file, and the Action to pass.

.github/workflows/build-docker.yml
IshaanDesai, line 1: We need to keep this CI.

.github/workflows/check-links.yml
IshaanDesai, line 1: Same as other CI related files. Let us keep this file to keep the check.

.github/workflows/check-markdown.yml
IshaanDesai, line 1: Same as other CI related files. Let us keep this file to keep the check.

.github/workflows/check-pep8.yml
IshaanDesai, line 1: Same as other CI related files. Let us keep this file to keep the check.

.github/workflows/run-tutorials.yml
IshaanDesai, line 1: Lets keep this file. It is okay if the tutorial CI fails for now. Eventually when we move the case to the precice/tutorials repository, the CI will pass again.

CHANGELOG.md
```diff
@@ -2,4 +2,4 @@
 
 ## current development
 
-* Forked initial version of this adapter from [`precice/fenics-adapter@v1.2.0`](https://github.com/precice/fenics-adapter/releases/tag/v1.2.0).
+* Forked initial version of this adapter from [`precice/fenics-adapter@v1.2.0`](https://github.com/precice/fenics-adapter/releases/tag/v1.2.0).
```
IshaanDesai: Not sure what the change is here. Please add a separate CHANGELOG entry describing the changes introduced in this pull request.

README.md
```diff
@@ -40,11 +41,11 @@ This should work out of the box, if all dependencies are installed correctly and
 
 Make sure to install the following dependencies:
 
-* [preCICE](https://github.com/precice/precice/wiki)
+* [preCICE v3](https://github.com/precice/precice/wiki)
```
IshaanDesai:
```suggestion
* [preCICE](https://precice.org/)
```

README.md
```diff
@@ -40,11 +41,11 @@ This should work out of the box, if all dependencies are installed correctly and
...
 * python3 (this adapter **only supports python3**)
 * [the python language bindings for preCICE](https://github.com/precice/python-bindings)
 * :construction: [FEniCSx](https://fenicsproject.org/) (with python interface, installed by default) (under construction refer to notes on FEniCSx below) :construction:
-* and scipy (`pip3 install scipy`)
+
```
IshaanDesai, line 48: We no longer need scipy?

README.md
```diff
@@ -91,3 +92,5 @@ If you are using FEniCSx, please also consider the information on [the official
 2021: For development of FEniCSx support, `precice/fenics-adapter@v1.2.0` was forked as `precice/fenicsx-adapter`. The required modifications were carried out by [Benjamin Rodenberg](https://www.cs.cit.tum.de/sccs/personen/benjamin-rodenberg/) and [Ishaan Desai](https://www.ipvs.uni-stuttgart.de/institute/team/Desai/).
 
 2023:  [Philip Hildebrand](https://github.com/PhilipHildebrand) updated the adapter to a first minimal working version (https://github.com/precice/fenicsx-adapter/pull/15) and contributed a first tutorial (https://github.com/precice/tutorials/pull/317) in the scope of his Bachelor's thesis ["Extending the FEniCSx Adapter for the Coupling Library preCICE"](https://mediatum.ub.tum.de/node?id=1706280) under supervision of [Benjamin Rodenberg](https://www.cs.cit.tum.de/sccs/personen/benjamin-rodenberg/) and [Ishaan Desai](https://www.ipvs.uni-stuttgart.de/institute/team/Desai/).
+
+2024:  [Eduardo Firvida](https://github.com/efirvida) updated the adapter to a minimal working version (https://github.com/precice/fenicsx-adapter/pull/26) with FSI and 3D support for preCICEv3. In the scope of his PhD Thesis.
```
IshaanDesai:
```suggestion
2024:  [Eduardo Firvida](https://github.com/efirvida) updated the adapter to support 3D FSI simulations and also made the adapter compatible with preCICE version 3 (https://github.com/precice/fenicsx-adapter/pull/26).
```

docs/ReleaseGuide.md
IshaanDesai, line 1: We need to keep the documentation.

fenicsxprecice/__init__.py
```diff
@@ -6,6 +6,7 @@
     warnings.warn("No FEniCSx installation found on system. Please check whether it is found correctly. "
                   "The FEniCSx adapter might not work as expected.\n\n")
 
-from .fenicsxprecice import Adapter
 from . import _version
+from .fenicsxprecice import Adapter
+
```
IshaanDesai: Is this reordering necessary?

fenicsxprecice/config.py
```diff
@@ -1,5 +1,4 @@
-"""
-This is the configuration module of fenicsxadapter
+"""This is the configuration module of fenicsxadapter
```
IshaanDesai: Let us keep the original formatting to maintain readability.

fenicsxprecice/config.py
```diff
@@ -8,8 +7,7 @@
 
 
 class Config:
-    """
-    Handles reading of config. parameters of the fenicsxadapter based on JSON
+    """Handles reading of config. parameters of the fenicsxadapter based on JSON
```
IshaanDesai: Same as above.

fenicsxprecice/config.py
```diff
@@ -25,8 +23,7 @@ def __init__(self, adapter_config_filename):
         self.read_json(adapter_config_filename)
 
     def read_json(self, adapter_config_filename):
-        """
-        Reads JSON adapter configuration file and saves the data to the respective instance attributes.
+        """Reads JSON adapter configuration file and saves the data to the respective instance attributes.
```
IshaanDesai: Same as above.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -1,44 +1,36 @@
-"""
-FEniCSx - preCICE Adapter. API to help users couple FEniCS with other solvers using the preCICE library.
-:raise ImportError: if PRECICE_ROOT is not defined
-"""
```
IshaanDesai, line 4: Let us keep these first documentation lines.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -1,44 +1,36 @@
...
 
 
-class Adapter:
-    """
-    This adapter class provides an interface to the preCICE coupling library for setting up a coupling case which has
-    FEniCSx as a participant for 2D problems.
-    The user can create and manage a dolfinx.UserExpression and/or dolfinx.PointSource at the coupling boundary.
-    Reading data from preCICE and writing data to preCICE is also managed via functions of this class.
-    If the user wants to perform implicit coupling then a steering mechanism for checkpointing is also provided.
+class Dimensions(NamedTuple):
```
IshaanDesai, line 16: Can you please add a short comment block stating what this class is doing?

fenicsxprecice/fenicsxprecice.py
```diff
@@ -1,44 +1,36 @@
...
 
-    For more information on setting up a coupling case using dolfinx.UserExpression at the coupling boundary please have
-    a look at this tutorial:
-    TODO
 
-    For more information on setting up a coupling case using dolfinx.PointSource at the coupling boundary please have a
-    look at this tutorial:
-    TODO
+class Adapter:
+    """This adapter class provides an interface to the preCICE v3 coupling library for setting up a
```
IshaanDesai, line 22:
```suggestion
    """
    This adapter class provides an interface to the preCICE v3 coupling library for setting up a
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -1,44 +1,36 @@
...
+    while ensuring seamless communication with other participants in the coupling process.
 
-    NOTE: dolfinx.PointSource use only works in serial
     """
 
-    def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.json'):
-        """
-        Constructor of Adapter class.
+    def __init__(self, mpi_comm, adapter_config_filename="precice-adapter-config.json"):
+        """Constructor of Adapter class.
```
IshaanDesai, line 33:
```suggestion
        """
        Constructor of Adapter class.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -47,30 +39,22 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
         adapter_config_filename : string
             Name of the JSON adapter configuration file (to be provided by the user)
         """
-
-        self._config = Config(adapter_config_filename)
-
-        # Setup up MPI communicator
         self._comm = mpi_comm
+        self._config = Config(adapter_config_filename)
```
IshaanDesai, line 43: No need for this reordering.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
         # Problem dimension in FEniCSx
         self._fenicsx_dims = None
 
-    def create_coupling_expression(self):
-        """
-        Creates a FEniCSx Expression in the form of an object of class GeneralInterpolationExpression or
-        ExactInterpolationExpression. The adapter will hold this object till the coupling is on going.
+    def interpolation_points_in_vector_space(self):
+        """Determine interpolation points in the vector space.
```
IshaanDesai, line 72:
```suggestion
        """
        Determine interpolation points in the vector space.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
-        raise Exception("PointSources are not implemented for the FEniCSx adapter.")
+        return unrolled_dofs.reshape(-1, bs), fs_coords[boundary_dofs]
 
     def read_data(self):
-        """
-        Read data from preCICE. Data is generated depending on the type of the read function (Scalar or Vector).
-        For a scalar read function the data is a numpy array with shape (N) where N = number of coupling vertices
-        For a vector read function the data is a numpy array with shape (N, D) where
-        N = number of coupling vertices and D = dimensions of FEniCSx setup
+        """Read data from preCICE.
```
IshaanDesai, line 94:
```suggestion
        """
        Read data from preCICE.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
-        read_data = None
-
-        if self._read_function_type is FunctionType.SCALAR:
-            read_data = self._interface.read_block_scalar_data(read_data_id, self._precice_vertex_ids)
-        elif self._read_function_type is FunctionType.VECTOR:
-            read_data = self._interface.read_block_vector_data(read_data_id, self._precice_vertex_ids)
-
-        read_data = {tuple(key): value for key, value in zip(self._fenicsx_vertices.get_coordinates(), read_data)}
-
+        np.ndarray
```
IshaanDesai, line 102:
```suggestion
        read_data : np.ndarray
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
+            The incoming data containing nodal data ordered according to _fenicsx_vertices
```
IshaanDesai, line 103:
```suggestion
            Data containing nodal data ordered according to _fenicsx_vertices
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
+        """
+        mesh_name = self._config.get_coupling_mesh_name()
+        data_name = self._config.get_read_data_name()
+        # For more information about readDara see and and time start here:
+        # https://precice.org/couple-your-code-porting-v2-3.html#add-relativereadtime-for-all-read-data-calls
```
IshaanDesai, line 108:
```suggestion
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
+        read_data = self._interface.read_data(
+            mesh_name, data_name, self._precice_vertex_ids, self.dt
+        )
         return copy.deepcopy(read_data)
 
     def write_data(self, write_function):
-        """
-        Writes data to preCICE. Depending on the dimensions of the simulation (2D-3D Coupling, 2D-2D coupling or
-        Scalar/Vector write function) write_data is first converted into a format needed for preCICE.
+        """Writes data to preCICE. Depending on the dimensions of the simulation.
```
IshaanDesai, line 115:
```suggestion
        """
        Writes data to preCICE. Depending on the dimensions of the simulation.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
-            raise Exception("write_function provided is neither VECTOR nor SCALAR type")
+        mesh_name = self._config.get_coupling_mesh_name()
+        write_data_name = self._config.get_write_data_name()
+        write_data = write_function.x.array[self.interface_dof]
+        self._interface.write_data(mesh_name, write_data_name, self._precice_vertex_ids, write_data)
 
     def initialize(self, coupling_subdomain, read_function_space=None, write_object=None):
-        """
-        Initializes the coupling and sets up the mesh where coupling happens in preCICE.
+        """Initializes the coupling and sets up the mesh where coupling happens in preCICE.
```
IshaanDesai, line 132:
```suggestion
        """
        Initializes the coupling and sets up the mesh where coupling happens in preCICE.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
 
         Parameters
         ----------
-        coupling_subdomain : Object of class dolfinx.cpp.mesh.SubDomain
-            SubDomain of mesh which is the physical coupling boundary.
-        read_function_space : Object of class dolfinx.functions.functionspace.FunctionSpace
+        coupling_subdomain : List
```
IshaanDesai, line 136:
```suggestion
        coupling_subdomain : list
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
+                self._config.get_coupling_mesh_name(),
+                self._config.get_write_data_name(),
+                self._precice_vertex_ids,
+                np.zeros(self._interface_dof_coords.shape),
+            )
+        self._interface.initialize()
+        return self.dt
+
+    def store_checkpoint(self, states: List):
+        """Defines an object of class SolverState which stores the current states of the variable and the time stamp."""
```
IshaanDesai, line 175:
```suggestion
        """
        Defines an object of class SolverState which stores the current states of the variable and the time stamp.
        """
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -84,280 +68,141 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
-        # making sure that the FEniCSx function provided by user is not directly accessed by the Adapter
-        assert (my_u != user_u)
-        self._checkpoint = SolverState(my_u, t, n)
-        self._interface.mark_action_fulfilled(self.action_write_iteration_checkpoint())
+        self._checkpoint = SolverState(states)
 
     def retrieve_checkpoint(self):
-        """
-        Resets the FEniCSx participant state to the state of the stored checkpoint.
+        """Resets the FEniCSx participant state to the state of the stored checkpoint.
```
```suggestion
        """
        Resets the FEniCSx participant state to the state of the stored checkpoint.
```

tests/README.md
IshaanDesai, line 1: Lets try to preserve the testing infrastructure as much as possible. I understand that tests related to old functionality are no longer relevant, but modifying existing tests to test the new functionality would be really helpful in catching potential bugs.

IshaanDesai (changes requested):
Thank you for refactoring your adapter design into the old design. I see now that most things in the old design are rendered unnecessary. The write function in the initialization can be removed as it no longer serves a purpose. I have some last suggestions, which I have written as comments. Once these are incorporated, we can finalize this pull request. Before merging, we should move the simulation setups to the tutorials repository. I can quickly port these once testing is done.

versioneer.py
BenjaminRodenberg, line 1: Can you revert the changes in this file? I think no changes should be necessary here.

fenicsxprecice/_version.py
BenjaminRodenberg, line 1: This file is generated by versioneer. Please revert the changes.

BenjaminRodenberg (commented):
Thanks a lot for all the work and the updates! I don't want to interfere too much here. Just two small comments on versioneer-related files. I would try to not mix these changes into the PR because it's not really related.

For some background: We are using versioneer in this repository to synchronize git-tags and the version numbers used in the `setup.py` respectively returned by `fenicsxadapter.__version__`.

BenjaminRodenberg:
To make this PR a bit easier to review and manage I reverted some changes in [fc02757](https://github.com/precice/fenicsx-adapter/pull/26/commits/fc027578cf8f5d8c5f200cdef617f80a680842c3). This should reduce the diff to some degree. @efirvida: If this interferes with your work, feel free to simply revert that commit (`git revert fc02757`). We can also use a different branch than `develop` to continue working on this PR.

I think we should generally try to factor some things out from this PR. Otherwise it is just too big. Some points that are part of this PR which we could factor out are minor changes like adding braces around `assert`s and some documentation updates. We should also try to fix the testing here in this PR in order to make sure we do not break anything. Additionally, this would also reduce the diff (currently there are many deletions due to the removal of the testing).

steffenger:
https://github.com/efirvida/fenicsx-adapter/blob/develop/fenicsxprecice/fenicsxprecice.py#L84 This version of the adapter only supports surface coupling now, right? It think it should be possible to make it more general with an additional argument in the initialization of the adapter. 
best regards

BenjaminRodenberg:
fyi: I added this PR to the milestone for the upcoming release (see https://github.com/precice/fenicsx-adapter/milestone/1). I think we should either close or merge this PR. In any case we should try to make sure we do not loose critical functionality provided in this PR in the process.

IshaanDesai:
Closing this PR because the adapter has advanced significantly, and merging it is no longer viable.

PR #28, opened 2024-12-12
Compatibility update for dolfinx 0.9.0 and preCICE v3

NiklasVin:
Updates the fenicsx-adapter such that it is compatible with dolfinx ~~0.8.0~~ 0.9.0 and preCICE v3.
The structure of the current state stays more or less the same but I am planning to add ideas and features from https://github.com/precice/fenicsx-adapter/pull/26.
The changes are also resolving https://github.com/precice/fenicsx-adapter/issues/17.
The changes also include the issues addressed in https://github.com/precice/fenicsx-adapter/pull/13. Unfortunately, I saw this PR too late and implemented more or less exactly the same

## Todos

- [x] Add entry in https://github.com/precice/fenicsx-adapter/blob/develop/CHANGELOG.md


BenjaminRodenberg:
Thank you for this update!

I changed 0.8.0 in the description to 0.9.0 since you also mention 0.9.0 in the title of this PR.

I will close #13 now assuming that the PR is outdated and superseeded by this one. I think any valuable changes in #13 should be moved over here if possible.

Please also update the README and mention FEniCSx Version 0.9.0 as dependency.

BenjaminRodenberg:
We should merge #26 such that the contribution of @efirvida shows up in the history. I see the danger that we move ideas from #26 into this PR and then #26 will never be merged into the history.

I don't have a very good overview here but I see #28 as a maintainment PR for compatibility with FEniCSx 0.9.0 (also updating the partitioned heat equation tutorial) and #26 as a feature PR (adding the perpendicular flap).

I see two options (depends on how crucial the ideas contributed by #26 for #28):

1) First merge #26 then #28: If merging #26 into `develop` leads to a non-working state I don't see this as a big problem as long as merging #28 will fix this situation.
2) First merge #28 then #26: If #28 is close to a working state that we can merge we can merge #28 without adding any ideas from #26 here. We could integrate the features provided via #26 at a later point in time.

fenicsxprecice/adapter_core.py
```diff
@@ -2,7 +2,8 @@
 This module consists of helper functions used in the Adapter class. Names of the functions are self explanatory
 """
 
-from dolfinx.fem import FunctionSpace, Function
+# from dolfinx.fem import FunctionSpace, Function
+from dolfinx import fem, geometry
```
BenjaminRodenberg:
```suggestion
from dolfinx import fem, geometry
```

fenicsxprecice/adapter_core.py
```diff
@@ -85,49 +86,48 @@ def determine_function_type(input_obj):
         raise Exception("Error determining type of given dolfin FunctionSpace")
 
 
-def convert_fenicsx_to_precice(fenicsx_function, local_ids):
+def convert_fenicsx_to_precice_coordinateBased(fenicsx_function, local_coords):
```
BenjaminRodenberg:
```suggestion
def convert_fenicsx_to_precice_coordinate_based(fenicsx_function, local_coords):
```

Is this renaming really necessary? If the new name is more appropriate we can do the rename.

Note that the main reason why I do not want to rename is that renaming also requires an update for the unit tests.

fenicsxprecice/expression_core.py
```diff
@@ -134,6 +134,10 @@ def lstsq_interp(x, y, w): return w[0] * x ** 2 + w[1] * y ** 2 + w[2] * x * y +
         # compute remaining error
         res = data - lstsq_interp(coords_x, coords_y, w)
         # add RBF for error
+
+        # TODO Add at a later point RBFInterpolator, because Rbf is deprecated
+        # rbf_interp = RBFInterpolator(np.column_stack((coords_x, coords_y)), res)
+        # return lambda x, y: rbf_interp(np.array([x, y]).T) + lstsq_interp(x, y, w)
```
BenjaminRodenberg: Open an issue and link here?

fenicsxprecice/fenicsxprecice.py
```diff
@@ -7,7 +7,7 @@
 from .config import Config
 import logging
 import precice
-from .adapter_core import FunctionType, determine_function_type, convert_fenicsx_to_precice, get_fenicsx_vertices, CouplingMode, Vertices
+from .adapter_core import FunctionType, determine_function_type, get_fenicsx_vertices, CouplingMode, Vertices, convert_fenicsx_to_precice_coordinateBased
```
```suggestion
from .adapter_core import FunctionType, determine_function_type, convert_fenicsx_to_precice_coordinateBased, get_fenicsx_vertices, CouplingMode, Vertices
```
Makes diff easier to read.

BenjaminRodenberg (commented):
I updated the tests and CI to work with this PR. Also added some minor comments. This is not a complete review.

I think it's already good news that the partitioned heat conduction test in the CI works :tada: 

**edit:** at least it worked for [a2c6c85](https://github.com/precice/fenicsx-adapter/pull/28/commits/a2c6c85a9edd51528bd12d8a90ecf6676917b5af). Not sure why I does not work anymore since the changes are unrelated.

fenicsxprecice/adapter_core.py
```diff
@@ -159,10 +159,9 @@ def get_fenicsx_vertices(function_space, coupling_subdomain, dims):
 
     # Get coordinates and IDs of all vertices of the mesh which lie on the coupling boundary.
     try:
-        on_subdomain = coupling_subdomain(mesh.geometry.x.T)
-        ids, = np.where(on_subdomain)
+        ids = fem.locate_dofs_geometrical(function_space, coupling_subdomain)
```
IshaanDesai, line 161: @steffenger is trying to run his simulation setup with this PR branch and he got the error:

```bash
I am participant "Fenicsx"
Caught the following exception in the detection of the coupling subdomain:
{e}
Falling back to old, point-wise method.
Traceback (most recent call last):
  File "/home/USADR/ac135212/spack/var/spack/environments/precice-env/.spack-env/view/lib/python3.12/site-packages/fenicsxprecice/adapter_[core.py](http://core.py/)", line 162, in get_fenicsx_vertices
    ids = fem.locate_dofs_geometrical(function_space, coupling_subdomain)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/USADR/ac135212/spack/var/spack/environments/precice-env/.spack-env/view/lib/python3.12/site-packages/dolfinx/fem/[bcs.py](http://bcs.py/)", line 51, in locate_dofs_geometrical
    return _cpp.fem.locate_dofs_geometrical(V._cpp_object, marker)  # type: ignore
```
He estimates that using the function `dfx.fem.locate_dofs_topological(...)` instead could potentially help. Maybe worth a try?

steffenger: https://github.com/NiklasVin/fenicsx-adapter/blob/adapter-update/fenicsxprecice/adapter_core.py#L175
I think my problem was related to my 3D mesh, since the adapter only handles 2d out of the box. I guess using `locate_dofs_geometrical` should be fine as long as your "marker function" can handle the coupling and there are no specific meshtags involved.

fenicsxprecice/fenicsxprecice.py
NiklasVin: I will change that

fenicsxprecice/adapter_core.py
NiklasVin: No, the renaming isn't necessary. I will change that.

NiklasVin:
@BenjaminRodenberg I think the second of the two options you suggested [here](https://github.com/precice/fenicsx-adapter/pull/28#issuecomment-2662289885) is better to handle after skimming #26 as it relies on the idea that the read and written data is the actual data of the dofs. This is why neither `expression_core.py` nor `adapter_core.py` are necessary in #26. The implementation of this PR doesn't assume this and if #26 is merged first, the two files are deleted and then added again what would make this PR even larger and less clear what has changed in `adapter_core.py`. Furthermore, the feature to support 3D cases from #26 should be easily addable on top of this PR.

fenicsxprecice/expression_core.py
BenjaminRodenberg:
```suggestion
        # TODO Add at a later point RBFInterpolator, because Rbf is deprecated
        # see https://github.com/precice/fenicsx-adapter/issues/29
        # rbf_interp = RBFInterpolator(np.column_stack((coords_x, coords_y)), res)
        # return lambda x, y: rbf_interp(np.array([x, y]).T) + lstsq_interp(x, y, w)
```

BenjaminRodenberg (approved):
I ran the tutorial and looks good. There are still some problems with errors and iteration count but let's deal with this at a later point in time.

I also fixed some tests.

Note that running the tutorial does not work out of the box since you need to setup the venv manually (remove `fenicsxprecice` from `requirements.txt` and install it with `pip3 install ../../../.`). But I also do not think we need this since the tutorial will be moved somewhere else at a later point in time.

BenjaminRodenberg:
What worries me a bit: From my perspective https://github.com/precice/fenicsx-adapter/pull/28/commits/bf5eda2469855bc283b7021d978277fdd9efa76d is necessary that the tests pass. At least locally I get errors if I do not apply this fix. However, the CI passes apparently for [c40fb25](https://github.com/precice/fenicsx-adapter/pull/28/commits/c40fb253047ab372d1adb2765835cee8edce6d67). We should investigate this more closely in the future since this means that our tests do not fail as expected.

setup.py
```diff
@@ -39,7 +39,7 @@
       author_email='info@precice.org',
       license='LGPL-3.0',
       packages=['fenicsxprecice'],
-      install_requires=['pyprecice~=2.0.0', 'scipy', 'numpy>=1.13.3', 'mpi4py'],
+      install_requires=['pyprecice~=3.0.0', 'scipy<1.15.0', 'numpy>=1.13.3', 'mpi4py'],
```
BenjaminRodenberg:
```suggestion
      install_requires=['pyprecice~=3.0', 'scipy<1.15.0', 'numpy>=1.13.3', 'mpi4py'],
```
Using `pyprecice~=3.0.0` would only allow `3.0` releases and not `3.1` etc. See https://stackoverflow.com/questions/39590187/in-requirements-txt-what-does-tilde-equals-mean.

IshaanDesai (approved):
I went through the adapter code changes, and they look reasonable :+1: I did not test the tutorial case as others have already done so. Do we want to merge the partitioned-heat-conduction case files?

BenjaminRodenberg:
> Do we want to merge the partitioned-heat-conduction case files?

Definitely! The tutorial structure also needs an update. The diff is a bit hard to read but I checked it again and there are no files that have been added accidentally.

I will merge this PR now.

PR #33, opened 2025-03-12
Install FEniCSx adapter in venv for running tutorials

BenjaminRodenberg:
## Main changes of this PR

Uses venv to install FEniCSx adapter

Follows the recommended approach for installing packages in python.

### DIFF
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -37,6 +37,9 @@ jobs:
           python3 -m venv .venv
           . .venv/bin/activate
           pip3 install ../../..
+      - name: Remove fenicsxadapter from requirements to avoid overwriting develop version
+        run: |
+          sed -i '/fenicsxprecice/d' tutorials/partitioned-heat-conduction/solver-fenicsx/requirements.txt
       # Tutorial is currently hosted under https://github.com/precice/fenicsx-adapter, should be moved to https://github.com/precice/tutorials when working. See https://github.com/precice/tutorials/issues/491
       # - name: Get tutorials
       #   run: git clone -b develop https://github.com/precice/tutorials.git
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -27,13 +27,13 @@ jobs:
         run: python3 -c "import dolfinx; print('FEniCS-X version '+dolfinx.__version__)"
       - name: Create venv for Dirichlet participants and install adapter
         run: |
-          cd tutorials/partitioned-heat-conduction/dirichlet-fenics
+          cd tutorials/partitioned-heat-conduction/dirichlet-fenicsx
           python3 -m venv .venv
           . .venv/bin/activate
           pip3 install ../../..
       - name: Create venv for Neumann participants and install adapter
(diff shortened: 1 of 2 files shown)


NiklasVin:
You could try to set `mpi4py<4` in `setup.py`? Maybe it's the same issue as here: https://fenicsproject.discourse.group/t/no-attribute-qualname-due-to-mpi4py-4-0-0/15375?

BenjaminRodenberg:
> You could try to set `mpi4py<4` in `setup.py`? Maybe it's the same issue as here: https://fenicsproject.discourse.group/t/no-attribute-qualname-due-to-mpi4py-4-0-0/15375?

Thanks for the hint! I will try this out. I'm a bit surprised that this did not explode in earlier runs (e.g. https://github.com/precice/fenicsx-adapter/commit/5d8b5536b85eb71dd9b85beb177c1ee18b74411d). But for some reason we installed mpi4py 3.x in https://github.com/precice/fenicsx-adapter/commit/5d8b5536b85eb71dd9b85beb177c1ee18b74411d and now we install mpi4py 4.x.

IshaanDesai (approved):

.github/workflows/run-tutorials.yml
```diff
@@ -25,8 +25,21 @@ jobs:
...
       - name: Check FEniCSx version
         run: python3 -c "import dolfinx; print('FEniCS-X version '+dolfinx.__version__)"
-      - name: Install adapter
-        run:  pip3 install --user .
+      - name: Create venv for Dirichlet participants and install adapter
+        run: |
+          cd tutorials/partitioned-heat-conduction/dirichlet-fenicsx
+          python3 -m venv .venv
+          . .venv/bin/activate
+          pip3 install ../../..
```
MakisH, line 33: An alternative to this would be to add the venv setup in the `run.sh`, as we do in the rest of the Python-related tutorials. Any reason to do this in the test environment here?

BenjaminRodenberg: I followed this approach because we need to install the latest version of the adapter before running the tutorial. If we want to use the `run.sh` from the tutorial file we would have to somehow inject a pointer to the local version of the repo in the CI pipeline into the `requirements.txt` before executing the `run.sh` script. (not sure if this is even possible) I think it's easier and clearer to just prepare the `venv` outside of the `run.sh` script and then use the `run.sh` without modifications.

BenjaminRodenberg: I follow the same procedure here for the same reason: https://github.com/precice/fenics-adapter/pull/198/files

MakisH: Ah, I see, same issue as with https://github.com/precice/tutorials/issues/584. I think it is possible to propagate the version, but indeed rather ugly.

But you still keep the venv mechanism in the tutorial as well, right? The user does not need to manually install the adapter, correct?

BenjaminRodenberg: Yes. The `run.sh` stays the same as always. Still needs some updates but we can do this when integrating into `precice/tutorials`

NiklasVin (approved):

PR #34, opened 2025-03-12
Support to communicate multiple data fields

### LINKED ISSUE #31: Support to read and write multiple data fields
This issue was already adressed for the fenics-adapter https://github.com/precice/fenics-adapter/issues/65. Allow the user to initialise the adapter also with dictionaries of different data or functionspaces could be a good solution, as described by @IshaanDesai in https://github.com/precice/fenics-adapter/issues/65#issuecomment-1407710576.

NiklasVin:
This pull request provides a first realisation of communicating multiple data fields as suggested in #31 and closes #31. The changes also affect the API of the fenicsx-adapter the following way:
* `create_coupling_expression()` → `create_coupling_expression(mesh_name)`
* `read_data(dt)` → `read_data(dt, mesh_name)`
* `write_data(write_function)` → `write_data(write_function, mesh_name)`
* `initialize(coupling_subdomain, read_function_space=None, write_object=None)` now expects a dictionary with entries of the following structure: `mesh_name: [coupling_subdomain, read_function_space, write_object]`

~Furthermore `fenicsxprecice` has a member `Meshes` which is an enum type and holds members with the same name as specified in the config files, i.e., if one participant defines a mesh called `Left` and you want to e.g. write values from this mesh, the implementation expects the user to call `write_data(write_function, fenicsxprecice.Meshes.Left)`. All functions that require `mesh_name` must use **fenicsxprecice.Meshes**.~

The structure of the participant config file also changed to support multiple meshes.

I also provided a minimal working example which (hopefully) makes it clear how to use this feature exactly (see `tutorials/multiple-fields`.


tutorials/partitioned-heat-conduction/precice-config.xml
```diff
@@ -10,37 +10,37 @@
   <data:scalar name="Temperature" waveform-degree="1" />
   <data:scalar name="Heat-Flux" waveform-degree="1" />
 
-  <mesh name="Dirichlet-Mesh" dimensions="2">
+  <mesh name="Dirichlet_Mesh" dimensions="2">
```
BenjaminRodenberg: Why did you change this to `Dirichlet_Mesh` etc?

tutorials/multiple-fields/precice-config.xml
```diff
@@ -0,0 +1,124 @@
...
+      enabled="true" />
+  </log>
+
+  <data:scalar name="LeftIn" waveform-degree="1" />
+  <data:scalar name="LeftOut" waveform-degree="1" />
+  <data:scalar name="RightIn" waveform-degree="1" />
+  <data:scalar name="RightOut" waveform-degree="1" />
+
+
+  <mesh name="Left1" dimensions="2">
```
BenjaminRodenberg:
```suggestion
  <mesh name="LeftOne" dimensions="2">
```
Also similar at some other places. Compare to the solverdummy where we use `SolverOne` and not `Solver1`

tutorials/multiple-fields/precice-adapter-config-R.json
```diff
@@ -0,0 +1,14 @@
+{
+  "participant_name": "Right",
+  "config_file_name": "precice-config.xml",
+  "interfaces": {
+    "Right1": {
+      "write_data_name": "LeftIn",
+      "read_data_name": "LeftOut"
+      },
```
BenjaminRodenberg: Does this mean that every mesh must have exactly one write and one read data? Is this restriction necessary?

Ideally we can implement the most general case where a mesh can have any number of write and read data (even none). But if this is not possible and working for the use-case of @steffenger without a lot of effort: We can also live with this restriction for now.

fenicsxprecice/config.py
```diff
@@ -39,17 +39,20 @@ def read_json(self, adapter_config_filename):
         data = json.load(read_file)
         self._config_file_name = os.path.join(folder, data["config_file_name"])
         self._participant_name = data["participant_name"]
-        self._coupling_mesh_name = data["interface"]["coupling_mesh_name"]
+        self._coupling_mesh_names = list(data["interfaces"].keys())
```
BenjaminRodenberg: Maybe we need to use something more general here: We could just parse the dictionary structure from the config and store it here as `self._meshes` then we could navigate to the read/write data by calling, for example, `self._meshes['Right1']['read_data']['Temperature']`.

fenicsxprecice/config.py
```diff
@@ -59,11 +62,25 @@ def get_config_file_name(self):
...
-        return self._read_data_name
+    def get_read_data_name(self, mesh_name):
+        """
+        Parameters
+        ----------
+        mesh_name : fenicsxprecice.Meshes member
+            Member of the enum fenicsxprecoce.Meshes
+
+        """
+        return self._read_data_names[mesh_name.name]
```
BenjaminRodenberg: Is this function really needed? I think the users should know what data they want to read. Using the name of the data (`str`) as a unique id that users directly provide to `read_data` and `write_data` should not be too complicated.

I assume the existence of this API is historic: Previously preCICE used lots of ids (`int`) to identify meshes and data. These ids had to be retrieved and stored during the initialization phase. Nowadays (preCICE >= 3) we simply use the mesh/data name to identify the meshes and data.

tutorials/multiple-fields/left.py
```diff
@@ -0,0 +1,64 @@
...
+coupling_expression1 = precice.create_coupling_expression(precice.Meshes.Left1)
+coupling_expression2 = precice.create_coupling_expression(precice.Meshes.Left2)
+
+while precice.is_coupling_ongoing():
+
+    if precice.requires_writing_checkpoint():
+        precice.store_checkpoint(uD, 0, 0)
+
+    read_data1 = precice.read_data(0, precice.Meshes.Left1)
+    read_data2 = precice.read_data(0, precice.Meshes.Left2)
```
BenjaminRodenberg: I think we can redesign this API in a way closer to the preCICE python API (compare [here](https://github.com/precice/python-bindings/blob/b4b9fe1a9f1ff5d699292ad9b55b8976a1778de0/examples/solverdummy/solverdummy.py#L65C39-L65C64))

```suggestion
    read_data1 = precice.read_data(0, precice.Meshes.Left1, precice.Meshes.LeftIn)
    read_data2 = precice.read_data(0, precice.Meshes.Left2, precice.Meshes.RightOut)
```

similar for write data.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -82,27 +85,35 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
...
         # Determine type of coupling in initialization
-        self._coupling_type = None
+        self._coupling_types = {}
 
         # Problem dimension in FEniCSx
         self._fenicsx_dims = None
         self._empty_rank = True
 
-    def create_coupling_expression(self):
+    def create_coupling_expression(self, mesh_name):
```
BenjaminRodenberg, line 91:
```suggestion
    def create_coupling_expression(self, mesh_name, read_data_name):
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -126,7 +137,7 @@ def update_coupling_expression(self, coupling_expression, data):
     def get_point_sources(self, data):
         raise Exception("PointSources are not implemented for the FEniCSx adapter.")
 
-    def read_data(self, dt):
+    def read_data(self, dt, mesh_name):
```
```suggestion
    def read_data(self, mesh_name, data_name, dt):
```
Let's add the `read_data_name` to allow multiple data per mesh and put `dt` to the end to make this consistent with the preCICE Python API.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -136,32 +147,42 @@ def read_data(self, dt):
...
+                    self._fenicsx_vertices[mesh_name].get_coordinates(),
+                    read_data)}
 
         else:
             pass
 
         return copy.deepcopy(read_data)
 
-    def write_data(self, write_function):
+    def write_data(self, write_function, mesh_name):
```
```suggestion
    def write_data(self, mesh_name, data_name, write_function):
```
see `read_data`

fenicsxprecice/fenicsxprecice.py
```diff
@@ -49,6 +50,8 @@ def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.jso
 
         self._config = Config(adapter_config_filename)
 
+        self.Meshes = Enum('Meshes', self._config.get_coupling_mesh_names())
+
```
BenjaminRodenberg: I think defining this Enum is pretty cool and also gives additional value to defining the mesh and data names initially by providing the `precice-adapter-config.json`.

1) At the start of the script the user has to provide a `precice-adapter-config.json` which defines the mesh names.
2) Later the user can provide the mesh/data names as strings to `read_data` and `write_data`. But it is also possible to use `Adapter.Meshes.MeshOne` instead.

However: I see this to some degree as an add-on and not as a must have. For simplicity I would rely on the user providing the correct string to `read_data` etc. we can later polish the API.

tutorials/multiple-fields/left.py
BenjaminRodenberg: Alternative idea for the API: `precice.LeftOne.LeftIn.read_data(dt)` or `precice.mesh('LeftOne').read_data('LeftIn').at(dt)`.

I think it depends a lot on what feels natural in FEniCSx. The API `precice.read_data(0, precice.Meshes.Left1, precice.Meshes.LeftIn)` is mainly inspired by the solverdummy.

BenjaminRodenberg (changes requested):
I just realized that I forgot to submit the review...

This is definitely already a very useful first prototype of the feature! Would be interesting if this already helps @steffenger.

fenicsxprecice/config.py
```diff
@@ -59,11 +62,25 @@ def get_config_file_name(self):
...
+        return self._coupling_mesh_names
 
-    def get_read_data_name(self):
-        return self._read_data_name
+    def get_read_data_name(self, mesh_name):
+        """
+        Parameters
+        ----------
+        mesh_name : fenicsxprecice.Meshes member
+            Member of the enum fenicsxprecoce.Meshes
```
BenjaminRodenberg:
```suggestion
            Member of the enum fenicsxprecice.Meshes
```

fenicsxprecice/config.py
```diff
@@ -59,11 +62,25 @@ def get_config_file_name(self):
...
+        return self._read_data_names[mesh_name.name]
 
-    def get_write_data_name(self):
-        return self._write_data_name
+    def get_write_data_name(self, mesh_name):
+        """
+        Parameters
+        ----------
+        mesh_name : fenicsxprecice.Meshes member
+            Member of the enum fenicsxprecoce.Meshes
```
BenjaminRodenberg:
```suggestion
            Member of the enum fenicsxprecice.Meshes
```

tutorials/partitioned-heat-conduction/precice-config.xml
NiklasVin: I changed it because due to the use of enums in `fenicsxprecice.Meshes`, names containing `-` cannot be converted into a member, since the enum would be named `Dirichlet-Mesh` but the minus is an arithmetic operator so python interprets this as an subtraction

tutorials/multiple-fields/precice-adapter-config-R.json
NiklasVin: If you are referring to if either can be optionally defined, then yes, this should be possible. I can modify the example regarding that.

tutorials/partitioned-heat-conduction/precice-config.xml
BenjaminRodenberg: I understand. Then we need a different solution on the enum side. We cannot change `"Dirichlet-Mesh"` to `Dirichlet_Mesh` because this restriction would also affect all the other tutorials.

You can also access the enum as follows: `meshes['Dirichlet-Mesh']`. `meshes.Dirichlet-Mesh`, unfortunately, does not work.

Another idea:

Generally you should be able to create enums where `enum.name` differs from the `enum.value`. You could use `dirichlet_mesh` as `enum.name` and `Dirichlet-Mesh` as `enum.value`.

`my_enums = Enum('enum',['enum-one','enum-two'])` does not work due to the reason you are mentioning (btw: This looks like a bug to me that would be worth reporting). `my_enums = Enum('enum',['enum_one','enum_two'])` works. If you look at `my_enums.enum_one.name` you get `'enum_one'` and the `my_enums.enum_one.value` is `1`.

However, I don't like the potential inconsistencies this approach introduces.

tutorials/multiple-fields/precice-adapter-config-R.json
steffenger: Hey @BenjaminRodenberg, 

I am also wondering about that restriction at the moment. If you have multiple data fields on the same mesh you get  counterintuitive configuration files. But in general, i think its good to give the user the functionality to use multiple meshes or marker functions if needed. How about using additional dictionaries for the read/write data structures?

`coupling_mesh = { 
"Mesh1" : [coupling_boundary, {"read_data1":  read_data1, "read_data2 : read_data2"}, {"write_data1" : write_data1, "write_data2" : write_data2} ]
}`

BenjaminRodenberg: I also think something going into this direction would be the final solution. However, we could implement the feature "as many meshes as you like with one read and write data per mesh" first and then generalize to "as many meshes as you like with arbitrarily many read and write data per mesh" in a second step. Does the first approach work for you already? There is some risk that we will not be able to implement the 100% general approach in the next few weeks because it is certainly more complicated.

steffenger: I was in the middle of trying it, but i am doing macro-micro coupling and i dont think the micro manager can handle multiple macro meshes (also when they're just formal), because it makes everything really complex:D.

Ps: what works for my coupling case at the moment is using the preCICE API, where i basically set the vertices and do the interpolation when needed by "hand".

IshaanDesai: Multiple meshes in a multiscale scenario is indeed not trivial. I would suggest first trying out this feature with a single mesh and multiple data fields in the read and write directions. Once this works for all desired scenarios, we can move to multiple meshes (single-scale and multiscale).

fenicsxprecice/config.py
NiklasVin: Since in this PR at most one read data name exists for each mesh and the preCICE API requires a data name we can omit the additional parameter for the data name. If we would have multiple read/write data for one mesh, then yes, the function is not needed anymore because the user must specify which data exactly should be written. In this PR, this avoids unnecessary parameters.

fenicsxprecice/fenicsxprecice.py
NiklasVin: Okay, in that case, I will remove the Enums for now and will have it in mind for later, when the adapter is in a better shape👍

fenicsxprecice/fenicsxprecice.py
NiklasVin: I would suggest that we either support multiple data per mesh later (in a seperate PR) or we open and merge a PR first in which the support for multiple data and a single mesh is implemented. I am working on a multiple mesh multiple data implementation already and I wouldn't suggest to mingle the two features in a single PR because it would get rather large. But I agree, changing the parameters to make it consistent with the preCICE API makes sense.

tutorials/partitioned-heat-conduction/precice-config.xml
NiklasVin: See [here](https://github.com/precice/fenicsx-adapter/pull/34#discussion_r2016325967).

fenicsxprecice/fenicsxprecice.py
BenjaminRodenberg: Fully agree to not do too many feature at the same time. If you think this PR is a good milestone on the way to the multi-data/multi-mesh PR: Feel free to wrap this one up first and then we build a second PR based on this one.

BenjaminRodenberg: In this case it should be `read_data(self, mesh_name, dt)` and the `data_name` is not necessary because there is a one-to-one relationship between `mesh_name` and (read)`data_name`.

Please don't forget to change the order of the arguments.

fenicsxprecice/fenicsxprecice.py
BenjaminRodenberg: `write_data(mesh_name, write_function)` ok? I think this is more consistent with `write_data(mesh_name, write_data_name, vertex_ids, write_data)` from the python bindings.

tutorials/partitioned-heat-conduction/neumann-fenicsx/precice-adapter-config-N.json
BenjaminRodenberg, line 1: Please keep the file here. You can follow the same approach like in the [FEniCS version of the tutorial](https://github.com/precice/tutorials/blob/4002c18e136e508b5a22242c8785fe06f4d06595/partitioned-heat-conduction/solver-fenics/heat.py#L102).

tutorials/multiple-fields/left.py
BenjaminRodenberg, line 1: I would suggest to directly follow a folder structure similar to the one in `partitioned-heat-conduction`. Should not be too difficult and it will only become more complicated in the future.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -170,139 +188,151 @@ def write_data(self, write_function):
...
+                write_function = write_object
+            # preCICE will use default zero values for initialization.
+            elif isinstance(write_object, fem.FunctionSpace):
+                write_function_space = write_object
+                write_function = None
+            elif write_object is None:
+                pass
+            else:
+                raise Exception(
+                    f"Given write object of {c_mesh} is neither of type dolfinx.functions.function.Function or dolfinx.functions.functionspace.FunctionSpace")
```
BenjaminRodenberg: This line has 159 characters and is clearly too long. But autopep8 passes. Let's not worry about this here but @NiklasVin: Can you please open an issue? I assume this is some kind of bug with autopep8 or our pipeline (not dramatic but we should document it).

fenicsxprecice/config.py
```diff
@@ -18,9 +18,8 @@ def __init__(self, adapter_config_filename):
 
         self._config_file_name = None
         self._participant_name = None
-        self._coupling_mesh_name = None
-        self._read_data_name = None
-        self._write_data_name = None
+        self._coupling_mesh_names = None
```
BenjaminRodenberg: Isn't this just `self._meshes.keys()`?

If yes: Better make it a function.

fenicsxprecice/config.py
```diff
@@ -59,11 +62,23 @@ def get_config_file_name(self):
     def get_participant_name(self):
         return self._participant_name
 
-    def get_coupling_mesh_name(self):
-        return self._coupling_mesh_name
+    def get_coupling_mesh_names(self):
+        return self._coupling_mesh_names
```
BenjaminRodenberg: Do you actually use this function? I think you can remove it.

tests/integration/test_fenicsxprecice.py
```diff
@@ -136,11 +136,11 @@ def right_boundary(x): return abs(x[0] - 1.0) < 10**-14
 
         precice = fenicsxprecice.Adapter(MPI.COMM_WORLD, self.dummy_config)
         precice._participant = Participant(None, None, None, None)
-        precice.initialize(right_boundary, self.scalar_V, self.scalar_function)
+        precice.initialize({"Dummy-Mesh": [right_boundary, self.scalar_V, self.scalar_function]})
```
BenjaminRodenberg, line 139: The dictionary is an appropriate data structure. We could also (not necessarily in this PR, because this will tunr it into a never-ending-story) introduce a class `CouplingMesh`. This would allow defining a constructor `CouplingMesh(name="Dummy-Mesh", domain=right_boundary, function_space=self.scalar_V, fenics_function=self.scalar_function)`. `precice.initialize` would then accept either a single `CouplingMesh` or a collection of `CouplingMesh`.

tests/integration/test_write_read.py
```diff
@@ -100,9 +100,9 @@ def return_dummy_data(n_points):
         precice = fenicsxprecice.Adapter(MPI.COMM_WORLD, self.dummy_config)
         precice._participant = Participant(None, None, None, None)
         precice._read_data_name = self.fake_data_name
-        precice.initialize(right_boundary, self.scalar_V)
+        precice.initialize({"Dummy-Mesh": [right_boundary, self.scalar_V, None]})
```
BenjaminRodenberg, line 103: I think this shows the problem of a dict containing lists following a certain pattern: We are forced to explicitly write `None` here for "formal" reasons. It would be much nicer to just write `CouplingMesh(name="Dummy-Mesh", domain=right_boundary, function_space=self.scalar_V)`.

tutorials/multiple-fields/left.py
```diff
@@ -0,0 +1,64 @@
...
+from dolfinx.fem.petsc import LinearProblem
+
+import fenicsxprecice
+
+domain = mesh.create_rectangle(
+    MPI.COMM_WORLD, [
+        numpy.asarray(
+            (0, 0)), numpy.asarray(
+                (1, 1))], [
+                    10, 10], mesh.CellType.quadrilateral)
```
BenjaminRodenberg:
```suggestion
domain = mesh.create_rectangle(
    MPI.COMM_WORLD, [
        numpy.asarray((0, 0)), 
        numpy.asarray((1, 1))
        ], 
    [10, 10], mesh.CellType.quadrilateral)
```
Line breaks are very odd in my opinion.

tutorials/multiple-fields/left.py
```diff
@@ -0,0 +1,64 @@
...
+domain2 = mesh.create_rectangle(
+    MPI.COMM_WORLD, [
+        numpy.asarray(
+            (0, 0)), numpy.asarray(
+                (1, 1))], [
+                    11, 11], mesh.CellType.quadrilateral)
+V = functionspace(domain, ("Lagrange", 1))
```
BenjaminRodenberg:
```suggestion
V1 = functionspace(domain, ("Lagrange", 1))
```
Because you are talking about `V2` below

tutorials/multiple-fields/left.py
```diff
@@ -0,0 +1,64 @@
...
+V2 = functionspace(domain2, ("Lagrange", 1))
+uD = fem.Function(V)
```
BenjaminRodenberg:
```suggestion
uD1 = fem.Function(V1)
```

tutorials/multiple-fields/right.py
```diff
@@ -0,0 +1,60 @@
...
+import numpy
+
+import fenicsxprecice
+
+domain = mesh.create_rectangle(
+    MPI.COMM_WORLD, [
+        numpy.asarray(
+            (1, 0)), numpy.asarray(
+                (2, 1))], [
+                    10, 10], mesh.CellType.quadrilateral)
```
BenjaminRodenberg: See `left.py`

tutorials/multiple-fields/right.py
```diff
@@ -0,0 +1,60 @@
...
+        numpy.asarray(
+            (1, 0)), numpy.asarray(
+                (2, 1))], [
+                    11, 11], mesh.CellType.quadrilateral)
+V = functionspace(domain, ("Lagrange", 1))
+V2 = functionspace(domain2, ("Lagrange", 1))
+uD = fem.Function(V)
+uD.interpolate(lambda x: 3 + x[0])
+uD2 = fem.Function(V2)
+uD2.interpolate(lambda x: 4 + x[0])
```
BenjaminRodenberg: See naming in `left.py`

BenjaminRodenberg (approved):
Looks good. Remaining comments are only cleanup and some refactoring.

Let's make this the final iteration. I don't think we should do another review and merge when you think that the comments are addressed in an appropriate way.

*edit:* I already factored out some unrelated changes from here and directly merged them into develop.

tests/integration/test_fenicsxprecice.py
NiklasVin: Yes, I also thought about it since the dict gets very unclear in the multiple meshes multiple data case. I will either include it in the PR for multiple meshes multiple data feature or in a seperate PR.

PR #35, opened 2025-03-12
Restrict to `mpi4py<4`

BenjaminRodenberg:
## Main changes of this PR

Restrict to `mpi4py<4` in `setup.py`

#33 fails (presumably due to mpi4py problem). In https://github.com/precice/fenics-adapter/pull/181 we implemented the identical fix.

## Author's checklist

* [x] I updated the changelog file `CHANGELOG.md` if there are user-observable changes since the last release. (skipping because this is a rather technical detail and we have not officially released the fenicsx adapter)

## Reviewers' checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setup(name='fenicsxprecice',
       author_email='info@precice.org',
       license='LGPL-3.0',
       packages=['fenicsxprecice'],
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
Will merge this now. Note that the failure of the test above will be fixed via #33. I also verified in #33 that the fix provided here actually fixes the installation procedure.

PR #39, opened 2025-04-26
Support to communicate multiple datafields on multiple meshes

### LINKED ISSUE #31: Support to read and write multiple data fields

NiklasVin:
## Main changes of this PR
* Introduction of the class `coupling_mesh`
* API changes for `read_data`, `write_data` and `initialize` needed to specify on which field of which mesh the data is communicated
* Tutorials and tests are modified according to the changes

#34 was a rather large PR. Therefore, all requirements from #31 were not added to #34.
With this PR, #31 can be closed.

## Author's checklist

## Reviewers' checklist


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## current development
 
+* Support to handle multiple data fields on one mesh [#39](https://github.com/precice/fenicsx-adapter/pull/39)
```
BenjaminRodenberg:
```suggestion
* Support to handle multiple data fields on one mesh. [#39](https://github.com/precice/fenicsx-adapter/pull/39)
```

tests/precice-adapter-config.json
```diff
@@ -3,8 +3,8 @@
   "config_file_name": "dummy.xml",
   "interfaces": {
     "Dummy-Mesh": {
-      "write_data_name": "Dummy-Write",
-      "read_data_name": "Dummy-Read",
+      "write_data_names": ["Dummy-Write"],
+      "read_data_names": ["Dummy-Read"],
```
BenjaminRodenberg, line 7:
```suggestion
      "write_data": ["Dummy-Write"],
      "read_data": ["Dummy-Read"],
```
I think this is a topic for another PR but I think we could just write `read_data`/`write_data` and then have a list of one or more (unique) names. I don't think we really need the suffix `_names`. Can you open an issue?

As far as I know `json` does not support unordered lists of unique objects ( = sets ). So we have to go with lists. In this case it would be reasonable to check the user input for duplicated names. If a duplicate occurs we should raise an error/warning.

fenicsxprecice/config.py
```diff
@@ -40,16 +40,16 @@ def read_json(self, adapter_config_filename):
...
-                self._meshes[mesh_name]["write_data_name"] = None
+                self._meshes[mesh_name]["write_data_names"] = None
 
             try:
-                self._meshes[mesh_name]["read_data_name"] = data["interfaces"][mesh_name]["read_data_name"]
+                self._meshes[mesh_name]["read_data_names"] = data["interfaces"][mesh_name]["read_data_names"]
             except KeyError:
                 # not required for one-way coupling, if this participant writes data
-                self._meshes[mesh_name]["read_data_name"] = None
+                self._meshes[mesh_name]["read_data_names"] = None
```
BenjaminRodenberg: Alternatively: Just design the constructor of `Mesh` such that you can write

```
mesh_data = data["interfaces"][mesh_name]
self._meshes.append(Mesh(mesh_name, mesh_data))
```

This would also allow you to check e.g. uniqueness of the data names inside the constructor and remove the `try...catch` blocks here.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -198,36 +198,78 @@ def write_data(self, mesh_name, write_function):
...
         write_data = convert_fenicsx_to_precice(write_function, self._fenicsx_vertices[mesh_name].get_coordinates())
         self._participant.write_data(
             mesh_name,
-            self._config.get_write_data_name(mesh_name),
+            write_data_name,
             self._precice_vertex_ids[mesh_name],
             write_data
         )
 
+    def validate_function_space(self, function_objects, mesh_name, checkIfFunction):
```
BenjaminRodenberg, line 215: Is this a refactoring that you could also merge into develop outside of this PR to make things a less complicated?

fenicsxprecice/fenicsxprecice.py
```diff
@@ -198,36 +198,78 @@ def write_data(self, mesh_name, write_function):
...
-            coupling_subdomain: Object of class dolfinx.cpp.mesh.SubDomain.
-                It is the subdomain of the mesh which is the physical coupling boundary.\n
-            read_function_space: Object of class dolfinx.functions.functionspace.FunctionSpace.
-                The function space on which the read function lives. If set to None then the adapter assumes that this
-                participant is a write-only participant.\n
-            write_object: Object of class dolfinx.functions.functionspace.FunctionSpace / dolfinx.functions.function.Function.
-                The Function space on which the write function lives or FEniCSx function related to the quantity to be written
-                by FEniCSx during each coupling iteration. If set to None then the adapter assumes that this participant is
-                a read-only participant.
+        coupling_meshes: A list of coupling meshes of the class Mesh.
```
```suggestion
        coupling_meshes: A list of coupling meshes of the class CouplingMesh.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -198,36 +198,78 @@ def write_data(self, mesh_name, write_function):
...
+                if checkIfFunction:
+                    raise Exception("A given object in {} is neither of type dolfinx.functions.function.Function or "
+                                    "dolfinx.functions.functionspace.FunctionSpace".format(mesh_name))
+                else:
+                    raise Exception(
+                        "A given object of {} is not of type dolfinx.functions.functionspace.FunctionSpace".format(mesh_name))
+
+        return function_space
+
     def initialize(self, coupling_meshes):
```
```suggestion
    def initialize(self, coupling_meshes: CouplingMesh):
```
Again: Maybe use typo annotations?

tutorials/multiple-fields/precice-config.xml
```diff
@@ -7,30 +7,32 @@
...
-  <data:scalar name="LeftIn" waveform-degree="1" />
-  <data:scalar name="LeftOut" waveform-degree="1" />
-  <data:scalar name="RightIn" waveform-degree="1" />
-  <data:scalar name="RightOut" waveform-degree="1" />
-
+  <data:scalar name="LeftInOne" waveform-degree="1" />
+  <data:scalar name="LeftInTwo" waveform-degree="1" />
+  <data:scalar name="LeftOutOne" waveform-degree="1" />
+  <data:scalar name="LeftOutTwo" waveform-degree="1" />
+  <data:scalar name="LeftOutThree" waveform-degree="1" />
```
BenjaminRodenberg, line 14: Why not create a new test here? I think the old one is good and the way you are modifying the old test (renaming + extension) makes this hard to review.

BenjaminRodenberg (changes requested):
Functionality-wise this PR looks good to me. I think there are still some refactoring that could go outside of this PR to simplify the review (feel free to directly merge refactorings into `develop`):

* Renamings (already exists)
* Factoring out `validate_function_space` (already exists)
* Introduce `classes` and use type annotations + mypy (open)

Note that last refactoring is a proposal. If you want to first merge this feature (time critical?) and do the refactoring later that's also fine.

fenicsxprecice/fenicsxprecice.py
NiklasVin: I decided to define this function because I need it two times, to check the given read and write objects (if their function spaces are equal). If I didn't define the function, there would be redundant code. This functionality was not necessary before, because then this condition was satisfied by default, so no refactoring

tutorials/multiple-fields/precice-config.xml
NiklasVin: I can undo the renaming, but I think the extension makes sense as it captures now every important feature that relates to #31. With this extended example, users can use this as a template for multiple mesh multiple data communication.

fenicsxprecice/config.py
NiklasVin: Ok, so you mean that it could make sense to introduce a class exclusively used in `config.py`? The `CouplingMesh` class I introduced was merely thought to make the initialization from the user's side clearer.

BenjaminRodenberg: My idea is to "upgrade" some part that you currently describe with a dict/json to a proper class. Main benefit here is that a class enforces a given structure why a dict/json does not. Additionally, you could ensure that it is only possible to create objects following this structure through an appropriate constructor. I hope this explanation is clear and does not add more confusion. We can also discuss this in person at our next meeting.

tutorials/multiple-fields/precice-config.xml
BenjaminRodenberg: I trust your judgement here. If you think the way how the test if written now this covers all cases then this is fine. Sometimes it is just easier to introduce a new test and keep the old one to make sure you test the old and new feature instead of modifying an existing one. But both approaches are valid in the end.

PR #46, opened 2025-07-27
Remove mpi4py restriction

### LINKED ISSUE #45: Restriction to mpi4py<4 still valid?
I was wondering if the restriction to mpi4py<4 is still valid, because in my case the adapter failed with mpi4py<4, throwing this error:

 terminate called after throwing an instance of 'nanobind::python_error
  what():  AttributeError: type object 'dolfinx.cpp.fem.CoordinateElement_float32' has no attribute __qualname__

However it worked with the following setup:

`pip list`

fenicsxprecice        0+untagged.227.g2e13673 
pyprecice             3.2.1 
mpi4py                4.1.0

IshaanDesai:
## Main changes of this PR

This PR removes the restriction of `mpi4py<4` because DOLFINx now works with the latest mpi4py. The partitioned heat conduction tutorial case has the restriction `numpy >1, <2`, which in turn requires mpi4py to be restricted to `<4`. Given this restriction, it is reasonable to move the mpi4py restriction to the tutorial, as it not a restriction from the side of the adapter.

## Author's checklist

## Reviewers' checklist

* [x] Does the changelog entry make sense? Is it formatted correctly?
* [x] Do you understand the code changes?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## current development
 
+* Remove version restriction on the dependency mpi4py. [#46](https://github.com/precice/fenicsx-adapter/pull/46)
 * Support to handle multiple data fields on one mesh. [#39](https://github.com/precice/fenicsx-adapter/pull/39)
 * Support communication of multiple data fields. [#34](https://github.com/precice/fenicsx-adapter/pull/34)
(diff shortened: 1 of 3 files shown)


NiklasVin (approved):
The changes make sense 👍

Issue #47, opened 2025-07-28
Suboptimal performance of the adapter


steffenger:
After @IshaanDesai made it possible to use the adapter for my case, I did some profiling with cProfile comparing the adapter coupling to the coupling with the preCICE API. 

**Coupling using the API via pyprecice**

```bash
Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1  256.812  256.812  610.928  610.928 lobule.py:1(<module>)
     9684  185.382    0.019  185.422    0.019 petsc.py:449(assemble_matrix_mat)
     8640  128.223    0.015  329.646    0.038 {method 'solve' of 'petsc4py.PETSc.SNES' objects}
    18325   10.432    0.001   10.451    0.001 assemble.py:186(_assemble_vector_array)
    17282    7.722    0.000   10.903    0.001 function.py:537(collapse)
1352325/103827    3.132    0.000    7.332    0.000 cell.py:253(__init__)
    18324    2.520    0.000    2.531    0.000 assemble.py:304(apply_lifting)
     1032    2.049    0.002    2.049    0.002 utils.py:235(write_function)
    34578    1.558    0.000    7.079    0.000 function.py:722(sub)
     9684    1.366    0.000    1.366    0.000 {method 'zeroEntries' of 'petsc4py.PETSc.Mat' objects}
    17282    0.844    0.000    0.844    0.000 function.py:442(_)
1768549/830765    0.557    0.000    5.881    0.000 cell.py:269(<genexpr>)
    51871    0.465    0.000    1.977    0.000 function.py:292(__init__)
  1352447    0.460    0.000    1.047    0.000 {built-in method _abc._abc_instancecheck}
  1696793    0.431    0.000    1.699    0.000 {built-in method builtins.isinstance}
   146906    0.393    0.000    0.393    0.000 la.py:254(array)
  1249120    0.389    0.000    0.612    0.000 ufl_type.py:36(__hash__)
   780330    0.379    0.000    0.481    0.000 ufl_type.py:40(__eq__)
...
```

**Coupling with adapter**

```bash
Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
1062331346/129606 1038.632    0.000 2790.306    0.022 copy.py:118(deepcopy)
849761280  439.846    0.000  439.846    0.000 {method '__deepcopy__' of 'numpy.generic' objects}
[...]

PR #48, opened 2025-08-20
Use just-in-time mapping to build custom interpolant in the adapter

NiklasVin:
## Main changes of this PR
- API for JIT-mapping feature (`read_data_at_coordinates()` and `write_data_at_coordinates()`)
- Added mwe `tutorials/jit-mapping`
- Changed the old preCICE calls which required `expression_core.py` with the new JIT-mapping functions in `tutorials/partitioned-heat-conduction`


fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
 
         return copy.deepcopy(read_data)
 
+    def read_data_at_coordinates(self, mesh_name, read_data_name, coordinates, dt):
+        """
+        Read data from preCICE at specified coordinates. This function uses the JIT-mapping feature.
```
```suggestion
        Read data from preCICE at specified coordinates. This function uses the just-in-time mapping of preCICE.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
...
+
+        Args:
```
```suggestion
        Parameters
        -----------------
```
Let us stick to one docstring style.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
...
+            mesh_name: Specifies for which mesh the data shall be read
+            read_data_name: Specifies the data name of the mesh
+            coordinates (list): List of coordinates where preCICE reads the data
+            dt: offset within time window
+
+        Returns:
```
```suggestion
        Returns
        -----------
```
Let us stick to one docstring style.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
 
         return copy.deepcopy(read_data)
 
+    def read_data_at_coordinates(self, mesh_name, read_data_name, coordinates, dt):
+        """
+        Read data from preCICE at specified coordinates. This function uses the JIT-mapping feature.
+
+        Args:
+            mesh_name: Specifies for which mesh the data shall be read
```
```suggestion
            mesh_name: Specifies for which mesh the data is read
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
...
+            read_data_name: Specifies the data name of the mesh
+            coordinates (list): List of coordinates where preCICE reads the data
```
```suggestion
            coordinates: List of coordinates where preCICE reads the data
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -180,6 +180,38 @@ def read_data(self, mesh_name, read_data_name, dt):
...
+            )
+
+            read_data = {
+                tuple(key): value for key, value in zip(coordinates, read_data)
+            }
+
+        else:
+            pass
+
+        return copy.deepcopy(read_data)
```
IshaanDesai, line 215: Can you please test if things work without doing this `deepcopy` operation? This operation is computationally very expensive, and becomes a significant burden when solving large problems.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -212,6 +244,21 @@ def write_data(self, mesh_name, write_data_name, write_function):
             write_data
         )
 
+    def write_data_at_coordinates(self, mesh_name, write_data_name, coordinates, write_function):
+        """
+        Writes data to preCICE at the given coordinates with the JIT-mapping feature
```
```suggestion
        Writes data to preCICE at the given coordinates with the just-in-time mapping of preCICE.
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -212,6 +244,21 @@ def write_data(self, mesh_name, write_data_name, write_function):
...
+
+        Args:
```
IshaanDesai: Same comment as above. Lets stick to one docstring style for the adapter.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -484,3 +531,21 @@ def requires_writing_checkpoint(self):
...
     def requires_reading_checkpoint(self):
         return self._participant.requires_reading_checkpoint()
+
+    def set_mesh_access_region(self, mesh_name, access_region):
+        """
+        Defines the access region for the JIT-mapping
+
+        Args:
+            mesh_name: The name of the mesh for which the access region is defined
+            access_region (list): A list of tuples defining the access region. Expects a list of 2 points.
```
IshaanDesai: Same comments as above.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -484,3 +531,21 @@ def requires_writing_checkpoint(self):
...
+        """
+        assert len(access_region) == 2  # should also be fine for 3d cases
+        assert len(access_region[0]) == len(access_region[1])
```
IshaanDesai: It is better to handle exceptions than add assertions.

IshaanDesai (changes requested):
This is a nice contribution, good job! Please have a look at my comments. Do you think there is value in keeping the jit-mapping tutorial? Perhaps for testing?

fenicsxprecice/fenicsxprecice.py
NiklasVin: At first glance, it should work without the `deepcopy` operation. I added this to keep it consistent with the `read_data()` function. Since https://github.com/precice/fenicsx-adapter/issues/47 refers to the issue with the `deepcopy`, I would prefer to open a PR removing all unnecessary copy operations at once and leave this one for consistency in the code. What do you think?

NiklasVin:
> Do you think there is value in keeping the jit-mapping tutorial? Perhaps for testing?

I think using it for testing could make sense, and as for the `multiple-fields` example, we could leave it in the repository as long as there are not many examples in which the jit-mapping is used to have a minimal working example as a reference?

IshaanDesai:
> > Do you think there is value in keeping the jit-mapping tutorial? Perhaps for testing?
> 
> I think using it for testing could make sense, and as for the `multiple-fields` example, we could leave it in the repository as long as there are not many examples in which the jit-mapping is used to have a minimal working example as a reference?

Sounds good. Lets keep it. But then it would make sense to add CI for this. We can add the CI in a separate pull request.

fenicsxprecice/fenicsxprecice.py
IshaanDesai: Yes, it makes sense to do the removal of the `deepcopy` operations in a separate PR. Let us leave it in here.

IshaanDesai (approved):
Looks good now, we can merge this.

While going over the changes it came to my mind if the way we expose the just-in-time mapping API to a FEniCS user is the right way to go. In principle we have the coordinates of the DoFs on the coupling boundary, so we could just use them internally. But maybe there is a use case for when data on the entire coupling boundary is not to be read or written. I will open an issue to discuss the API exposure.

PR #51, opened 2025-09-05
Add support for 3D coupling

### LINKED ISSUE #36: support for 3d models
In the current version, the FEniCSx adapter is only able to deal with 2d cases. however, in multiphysics coupling simulations such as FSI, the calculation models are usually 3 dimensions. so expanding the three-dimensional computational model will greatly enhance the software's practicality.

NiklasVin:
## Main changes of this PR

Add 3d support via just-in-time mapping

Closes #36 and is an alternative suggestion to support 3D cases to #26.

## Author's checklist

## Reviewers' checklist


fenicsxprecice/fenicsxprecice.py
```diff
@@ -183,6 +183,7 @@ def read_data(self, mesh_name, read_data_name, dt):
     def read_data_at_coordinates(self, mesh_name, read_data_name, coordinates, dt):
         """
         Read data from preCICE at specified coordinates. This function uses the just-in-time mapping of preCICE.
+        It can be used for 2D and 3D cases.
```
```suggestion
```
No need for this comment. By default a user should assume that the functions are usable for all dimensional scenarios that preCICE supports.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -249,6 +250,7 @@ def write_data(self, mesh_name, write_data_name, write_function):
     def write_data_at_coordinates(self, mesh_name, write_data_name, coordinates, write_function):
         """
         Writes data to preCICE at the given coordinates with the just-in-time mapping of preCICE.
+        It can be used for 2D and 3D cases.
```
IshaanDesai: Same as above.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -386,9 +388,13 @@ def initialize(self, coupling_meshes: list[CouplingMesh]):
...
 
             # Set up mesh in preCICE
-            self._precice_vertex_ids[mesh_name] = self._participant.set_mesh_vertices(
-                mesh_name, self._fenicsx_vertices[mesh_name].get_coordinates()[
-                    :, :2])  # give preCICE only 2D coordinates
+            if self._fenicsx_dims == 2:
+                self._precice_vertex_ids[mesh_name] = self._participant.set_mesh_vertices(
+                    mesh_name, self._fenicsx_vertices[mesh_name].get_coordinates()[
+                        :, :2])  # give preCICE only 2D coordinates
+            else:
```
IshaanDesai, line 393: Do we still need to handle the 2D case separately? Why can't we just hand over `self._fenicsx_vertices[mesh_name].get_coordinates()` to preCICE?

fenicsxprecice/fenicsxprecice.py
```diff
@@ -400,19 +406,17 @@ def initialize(self, coupling_meshes: list[CouplingMesh]):
...
-                        raise Exception(
-                            "preCICE requires you to write initial data. Please provide a write_function to initialize(...)")
-                    self.write_data(mesh_name, write_data_name, write_function)
+                if c_mesh.get_write_fields() is not None:
+                    for write_data_name in c_mesh.get_write_fields():
+                        write_function = c_mesh.get_write_fields()[write_data_name]
+                        if not isinstance(write_function, fem.Function):
+                            raise Exception(
+                                "preCICE requires you to write initial data. Please provide a write_function to initialize(...)")
+                        self.write_data(mesh_name, write_data_name, write_function)
```
IshaanDesai: Is this related to 3D cases? Or is it about checking if an appropriate write function is provided to the initialize routine?

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/errorcomputation.py
```diff
@@ -0,0 +1,21 @@
+from dolfinx import fem
+import numpy as np
+from mpi4py import MPI
+import ufl
+
+
+def compute_errors(u_approx, u_ref, total_error_tol=10 ** -4):
```
IshaanDesai: Can we have a stricter tolerance here? Is it possible to recover the analytical solution?

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/problem_setup.py
```diff
@@ -0,0 +1,113 @@
...
+import dolfinx.io.gmshio
+from dolfinx.mesh import DiagonalType, create_box
+import dolfinx.mesh
+from my_enums import DomainPart
+import numpy as np
+from mpi4py import MPI
+import gmsh
+
+
+def get_geometry(domain_part):
```
IshaanDesai, line 13: Can you please add comments to the domain filtering `if` to state which part of the 3D box domains are filtered in which conditions. Essentially adding comments to give the reader a bit of a sense of what the geometry is and how it is constructed.

IshaanDesai (changes requested):
Thanks for this important contribution! Please see my comments.

fenicsxprecice/fenicsxprecice.py
NiklasVin: It should be still necessary, because if a 2D case is defined in the preCICE config, preCICE expects 2D coordinates and throws an error as soon as the dimensions of the coordinates are not identical.

fenicsxprecice/fenicsxprecice.py
NiklasVin: No it is not related to 3D cases. I forgot to remove it from this PR - I just happen to see that this could potentially be a problem

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/errorcomputation.py
NiklasVin: I can change the tolerance for sure. But I don't know if I am able to recover the analytical solution; with some fine-tuning I could reduce the error but I am not sure by how much...

IshaanDesai: Okay. The question was more from a curiosity-driven perspective. If we find out the minimum tolerance that we can work with, that would be sufficient.

fenicsxprecice/fenicsxprecice.py
IshaanDesai: Okay. Lets add this via a different PR, or if you think it is trivial, just directly add it on the develop branch.

fenicsxprecice/fenicsxprecice.py
IshaanDesai: Okay, yes, makes sense.

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/errorcomputation.py
NiklasVin: I changed the tolerance and allowed 200 iterations per timestep since preCICE didn't converge with only 100. It reduced the error a bit but not by magnitudes.

IshaanDesai: That is good enough for now :+1:

IshaanDesai (approved):

PR #55, opened 2025-11-07
Add proper interpolation for 2D and 3D cases and support for MPI

### LINKED ISSUE #9: Enable Vector valued coupling

### LINKED ISSUE #10: Provide an option to configure the rbf-interpolation
The default rbf-iterpolation https://github.com/precice/fenicsx-adapter/blob/6db7f454330433ad745a8e08a0cd122937b5474e/fenicsxprecice/expression_core.py#L157 currently used in the adapter may lead to issues for some cases, see e.g. https://precice.discourse.group/t/problem-with-rbf-interpolation-in-the-fenics-x-adapter/1037. We should provide an option to configure the rbf parameters.

### LINKED ISSUE #22: Reimplement parallelization
In #15 we decided to remove parallelization to make the scope more realistic. There is already some work concerning parallelization in https://github.com/precice/fenicsx-adapter/pull/1/commits/26678c169022bfc1776f1fa2d8bad40e4fbf825b and the FEniCS adapter https://github.com/precice/fenics-adapter could also serve as a tempalte.

### LINKED ISSUE #29: Replace outdated `Rbf` with `RBFInterpolator`
The class Rbf that is used in `expression_core.py` is according to the [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Rbf.html) outdated and no longer maintained.
At some point it could make sense to replace the `Rbf` with `RBFInterpolator`.

NiklasVin:
## Main changes of this PR
* Instead of assigning dof values of a FEniCSx function directly, the changes made to the adapter now allow it to internally define the preCICE mesh such that it matches the interpolation points required by FEniCSx's `interpolate()` function. 
* The user can decide if they want to use the new internal interpolation method or if they want to define the preCICE mesh and the interpolation themselves. (cf. `CouplingBoundaryInterpolation` enum)
* The new interpolation has full MPI support
* Remove `coupling_expression`
The `coupling_expression` was to restrictive (only 2D and scalar valued functions allowed) and leaving the interpolation in the solvers is tedious.
Additionally, merging this PR closes #9, #10, #22 , #29. 
## Author's checklist

## Reviewers' checklist


fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
     cell_candidates = geometry.compute_collisions_points(bb_tree, local_coords)
     # Choose one of the cells that contains the point
     colliding_cells = geometry.compute_colliding_cells(mesh, cell_candidates, local_coords)
+
+    # generate from the already computed cell candidates the midpoint tree!
```
IshaanDesai:
```suggestion
    # generate the midpoint tree from the already computed cell candidates
```

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+            # point is outside domain, probably because of rounding
+            closest_cell_idx = geometry.compute_closest_entity(bb_tree, midpoint_tree, mesh, point)[0]
+            # change point such that it is in the function domain
+            # -> find midpoint to closest cell
+            closest_midpoint_cell = msh.compute_midpoints(mesh, mesh.topology.dim, np.array([closest_cell_idx]))[0]
+            # in each direction, COORDINATE_DIGITS defines on how much the point has been shifted
+            # -> move it in the direction of midpoint cell by at max. this amount, so, in each direction maximal 1e-COORDINATE_DIGITS
+            direction = closest_midpoint_cell - point
+            direction = np.sign(direction) * (10**(-COORDINATE_DIGITS))
+            new_point = point + direction
```
IshaanDesai:
```suggestion
            moved_point = point + direction
```

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+
+            cells.append(closest_cell_idx)
+            points.append(new_point)
```
IshaanDesai:
```suggestion
            points.append(moved_point)
```

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+    interpolation_coordinates = []
+    vec_len = function_space.dofmap.bs
+
+    def query_coordinates(x):
+        # to avoid UnboundLocalError, interpolation_coordinates is an array to which x is appended to
+        interpolation_coordinates.append(np.transpose(copy.deepcopy(x)))
+        # directly round and make each rounded coordinate unique to keep the code clean and to reduce memory consumption
+        interpolation_coordinates[-1] = round_unique_coordinates(interpolation_coordinates[-1])
+
+        return np.zeros((vec_len, x.shape[1]))
```
IshaanDesai:
```suggestion
        return np.zeros((vec_len, x.shape[1]))
```

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+    dofs_coupling = fem.locate_dofs_geometrical(function_space, coupling_boundary)
+    dofs_coupling_coordinates = function_space.tabulate_dof_coordinates()[dofs_coupling]
+    bb_tree = geometry.bb_tree(domain, domain.geometry.dim)
+    cell_candidates_local = geometry.compute_collisions_points(bb_tree, dofs_coupling_coordinates)
+    cell_candidates_local = geometry.compute_colliding_cells(domain, cell_candidates_local, dofs_coupling_coordinates)
+    # the cell candidates with local cell ids
+    cell_candidates_local = np.unique(cell_candidates_local.array)
+
+    # determine process-owned cells
+    index_map = domain.topology.index_map(domain.topology.dim)
```
IshaanDesai, line 239: Looks odd because `domain.topology` should have access to `dim` internally.

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+
+    # send interpolation points of complete boundary (interpolation_coordinates[0]) to higher ranks
+    # OR receive interpolation points that potentially need to be filtered out from lower ranks
+
+    # TODO add persistent communication for performance increase?
+
+    duplicate_coordinates_per_rank = {}  # save the filtered coordinates per rank to reduce communication volume
+
+    for source_rank in range(comm_rank):
+        interpolation_coordinates_from_source = comm.recv(source=source_rank)  # recv a set of tuples
```
IshaanDesai, line 269: Does this cover the scenario when a rank needs to send as well as receive? Or would this never be the case in the algorithm?

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+    return interpolation_coordinates, owned_and_candidate_cells_local, coordinates_to_send
+
+
+def interpolate_boundary_function(
+        read_values: dict,
+        function_type: FunctionType,
+        values_to_send: dict,
+        boundary_function: fem.Function,
+        boundary_cells: list,
+        comm: MPI.Comm,
```
IshaanDesai, line 300: Why not `MPI.COMM_WORLD`? Are we really using custom communicators?

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
-                    coords.append([v[0], v[1]])
-        ids = np.array(ids)
-        coords = np.array(coords)
-    return ids, coords
+            # vector valued function
+            # vector length is determined by getting one value of the values dict
+            vector_length = len(read_values[next(iter(read_values))])
+
+        # append filtered coordinates again to provide all necessary points for interpolation
+        for source_rank in range(comm.Get_rank()):
```
IshaanDesai, line 341: I am confused by this. Does this mean the `source_rank` here is the rank itself?

fenicsxprecice/adapter_core.py
```diff
@@ -122,58 +155,210 @@ def convert_fenicsx_to_precice(fenicsx_function, local_coords):
...
+            value = comm.recv(source=source_rank)
+            read_values.update(value)
+
+        # send the values other ranks need for interpolation
```
IshaanDesai:
```suggestion
        # send the values that other ranks need for interpolation
```

fenicsxprecice/fenicsxprecice.py
```diff
@@ -321,21 +305,27 @@ def validate_function_space(self, function_objects, mesh_name, check_if_conditio
...
 
-    def initialize(self, coupling_meshes: list[CouplingMesh]):
+    def initialize(self, coupling_meshes: list[CouplingMesh], precice_mesh=None):
         """
         Initializes the coupling and sets up the mesh where coupling happens in preCICE.
 
         Parameters
         ----------
         coupling_meshes: A list of coupling meshes of the class CouplingMesh.
+        precice_meshes: A list of dicts containing the definition of preCICE meshes for each CouplingMesh.
```
IshaanDesai: One more sentence describing what a "preCICE mesh definition" is would be good here.

fenicsxprecice/fenicsxprecice.py
```diff
@@ -381,25 +371,31 @@ def initialize(self, coupling_meshes: list[CouplingMesh]):
...
-            # this isnt a problem in update_coupling_expression, because in this function
-            # , the two first dimensions are extracted. Exactly what we want!
+
+            if self.boundary_proc_mode == CouplingBoundaryInterpolation.ADAPTER:
+                coords, cells, self._values_to_send[mesh_name] = get_fenicsx_interpolation_points(
+                    function_space, c_mesh.get_coupling_boundary(), self._comm)
+                ids = np.arange(len(coords))
+                self._interpolation_cells[mesh_name] = cells
+            else:
+                # precice_mesh = {ids : coordinates}
```
```suggestion
```

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/heat.py
```diff
@@ -69,6 +69,7 @@ def compute(self, u):
         L = fem.form(ufl.inner(ufl.grad(u), self.v) * ufl.dx)
         b = create_vector(L)
         assemble_vector(b, L)
+        b.ghostUpdate(addv=PETSc.InsertMode.ADD_VALUES, mode=PETSc.ScatterMode.REVERSE)
```
IshaanDesai, line 74: Can you add a comment on what the settings here are?

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/heat.py
```diff
@@ -180,10 +182,10 @@ def __call__(self, x):
 name_read = None
 if problem is ProblemType.DIRICHLET:
     name_read = "Temperature"
-    read_mesh = "Neumann-Mesh"
+    read_mesh = "Dirichlet-Mesh"
```
IshaanDesai, line 187: Why is this change required?

IshaanDesai (commented):
This is a major feature and upgrade of the adapter. Great work! I have some questions and comments, mainly to understand things. The code already looks very good.

fenicsxprecice/adapter_core.py
NiklasVin: For a 3d domain, you could also create an index map for e.g. the facets instead of the cells (so topological dimension minus one). That's why there is a parameter for the topological domain (if you are referring to that?).

fenicsxprecice/adapter_core.py
NiklasVin: Yes, in the long comment above I defined a rule how the ranks communicate. ranks from 1 to n-1 receive from the lower numbered ranks AND send to higher numbered ranks. So, rank 0 only sends and rank n only receives. After rank x received messages from ranks 0 to x-1, it sends to ranks x+1 to n. Therefore, after rank x received x messages the for-loop is finished and it continues to the "send loop".

fenicsxprecice/adapter_core.py
NiklasVin: `MPI.COMM_WORLD` is also of type `MPI.Comm` and I thought this might be more general as we don't know what the user wants to do with their MPI ranks (when invoking the fenicsxprecice constructor, we allow the user to pass any communicator, so this could then probably break the user's desired MPI setup internally, if I am not mistaken)

fenicsxprecice/adapter_core.py
NiklasVin: I am not sure what you mean exactly. In this loop, rank x receives the interpolation values it requires but which cannot be retrieved directly from preCICE (as this value at the specific coordinate is defined on another rank's preCICE mesh). This follows again the same "rule as above", namely that rank x gets it's values from ranks 0 to x-1, so the `source_ranks`.

tutorials/partitioned-heat-conduction-3d/solver-fenicsx/heat.py
NiklasVin: This change is required because before, this tutorial used jit-mapping and in this case, a different config was necessary

fenicsxprecice/adapter_core.py
IshaanDesai: Valid point :+1:

fenicsxprecice/adapter_core.py
IshaanDesai: Okay. I meant it more from the software perspective that it looks like you are passing the `dim` variable of `domain.topology` into a function, which is again of `domain.topology`. This is not too important, though, so let's leave it like it is if it works.

fenicsxprecice/adapter_core.py
IshaanDesai: Okay, makes sense.

fenicsxprecice/adapter_core.py
IshaanDesai: Okay, now it makes sense. I misunderstood the code before.

PR #61, opened 2025-12-14
Resolve issue with segmentation faults

### LINKED ISSUE #58: Segfaults for certain functionspaces
The following generate two function spaces that should effectively be vector fields, in which vectors have size 6.

```python
We = basix.ufl.quadrature_element(domain.basix_cell(), degree=0, scheme="default", value_shape=(6,))
W = fem.functionspace(domain, We)

V = fem.functionspace(domain, ("CG", 2, (6,)))
```

However, fenicsx treats W as scalar field, in which each "scalar" is a vector of size 6.

Upon calling `Adapter.write_data` from fenicsxprecice.py:216, it calls `determine_function_type` from adapter_core.py:58.
In there the function type is determined using the number of sub spaces via `input_fspace.num_sub_spaces`.

One option to mitigate this, might be by determining the function type via the function's value_size, i.e., `input_fspace.value_size`.

Due to this misinterpretation of the function space, most likely preCICE buffers are accessed incorrectly, leading to the Segfault.

NiklasVin:
## Main changes of this PR
Instead classifying the coupling functions with the property `num_function_spaces`, it is determined whether it is scalar- or vector-valued with the property `value_size`.

Closes #58 

## Author's checklist

## Reviewers' checklist

Co-authored-by: @Snapex2409


IshaanDesai:
@Snapex2409 can you please try out the changes in this PR for your case to make sure that the segmentation fault is resolved? Based on your feedback, we will be able to merge this.

PR #64, opened 2025-12-31
Add release workflow and update readme

NiklasVin:
## Main changes of this PR

This PR adds the release workflow and updates the readme according to the coming up release.

## Author's checklist

## Reviewers' checklist


IshaanDesai (approved):
The only way to test the release workflow is to merge this and prepare a release candidate. Then we will know if the preemptively set trusted publishing works.

PR #67, opened 2026-02-08
Contribution to Issue #66: Int based implementation

### LINKED ISSUE #66: Round Unique Coordinates yields different results on different machines
In `adapter_core.py:round_unique_coordinates`: 

Due to the implementation of `np.round`, which does not use true mathematical rounding but rather a specific algorithm (see [Numpy Doc)](https://numpy.org/doc/2.1/reference/generated/numpy.round.html), rounding close to machine precision is in accurate and depends on the underlying CPU-Architecture. This results in different final results on different machines.

```python
def round_unique_coordinates(coords, digit_cutoff):
    tmp = np.zeros_like(coords)
    np.round(coords, digit_cutoff, tmp)
    tmp = np.unique(tmp, axis=0)
    return tmp
```

I suggest 2 options:
- implement int based comparison, which splits coordinates into sections of ints
- implement oct-tree based filtering: minimum cell-size = digit_cutoff, all points which end up in the same cell are merged to a single entity

Snapex2409:
## Main changes of this PR

Implemented int based unique coordinates version.
Splits float values into reasonable sizes (10 digits) and coverts those to int64 (~16 Digits). 
Unique operation on int values does not need rounding anymore.

Resolves [Issue #66](https://github.com/precice/fenicsx-adapter/issues/66)

## Author's checklist

## Reviewers' checklist


IshaanDesai (approved):
Looks good. The tutorial also runs as expected 👍

IshaanDesai:
The CI fails due to redundant checks in the tutorial case, which will be fixed in another PR.

PR #78, opened 2026-05-08
Introduce cache for `convert_fenicsx_to_precice`

NiklasVin:
## Main changes of this PR

I added a cache for the evaluation of the fenicsx functions to improve performance for `write_data`.

The caching avoids recomputing the cell candidates necessary to evaluate the FEniCSx function.
Since `write_data_at_coordinates` requires the user to pass the coordinates, evaluating whether all coordinates passed to the function are the same as in the previous call, can get expensive, since a simple `==` comparison of the previous and the current coordinate array shouldn't be sufficient (if the ordering of the coordinates have changed, at least a set operation is necessary).
It is an option to give the user the possibility to not provide a list of coordinates for the `read_data_at_coordinates`, and use the last coordinates for the mapping. If this should be allowed, then I would rather revise the whole separation of `fenicsxprecice.py` and `adapter_core.py` and potentially change the adapter architecture since the majority of the `adapter_core.py` would definitely benefit from that (you can see that for instance for the `digit_cutoff` parameter that more or less all functions require, or `interpolate_boundary_function` also takes a very long list of parameters).

## Author's checklist

## Reviewers' checklist


fenicsxprecice/fenicsxprecice.py
IshaanDesai, line 1: Where is the user configuration for whether to use cached coordinates?

IshaanDesai (commented):
Let us do the design change separately.

fenicsxprecice/fenicsxprecice.py
NiklasVin: There is none. Since I only implemented it for the write method for static meshes, the cached coordinates will never change

IshaanDesai (approved):
Looks good 👍 feel free to squash and merge. Please remember to add a changelog entry.

PR #80, opened 2026-05-31
Several fixes and enhancements for parallel use

IshaanDesai:
## Main changes of this PR

This PR:

- adds profiling sections for performance analysis of the adapter components.
- fixes a potential bug in the interpolation when the adapter is used in parallel.
- replaces the blocking communication in the interpretation functionality with non-blocking communication to reduce runtime for parallel cases.

Run the adapter in parallel.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/fenicsxprecice/fenicsxprecice.py
+++ b/fenicsxprecice/fenicsxprecice.py
@@ -219,7 +219,7 @@ class Adapter:
         """
         if self._empty_rank:
             return
-        else: # to avoid unnecessary profiling sections for ranks that do not write data
+        else:  # to avoid unnecessary profiling sections for ranks that do not write data
             self.start_profiling_section("fenicsxprecice.write_data")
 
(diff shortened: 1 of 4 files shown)


fenicsxprecice/fenicsxprecice.py
```diff
@@ -38,7 +37,7 @@ class Adapter:
 
     def __init__(self, mpi_comm, adapter_config_filename='precice-adapter-config.json',
                  boundary_processing_mode=CouplingBoundaryInterpolation.ADAPTER,
-                 digit_cutoff=25):
+                 digit_cutoff=10):
```
NiklasVin: This shouldn't be necessary, right?

NiklasVin (changes requested):

fenicsxprecice/fenicsxprecice.py
IshaanDesai: Yes, correct. This is leftover from the latest runs. I will remove it.

